PR #15, opened 2018-04-08
Fixed an issue where Retrieve calls for a specific would fail, because the request was not being referenced

david-driscoll:
There is a small typo where the request is not being used an error is being throw from JavaScript strict mode (`paritionKey` is not defined)


edwin-huber:
Hi David, thanks for your contribution!

I'm just working through a couple of things, and will then be updating the project accordingly.
Please give me a couple days to get everything straight.

edwin-huber:
This has already been merged. :)

PR #25, opened 2018-05-25
BLOB shared key authentication

vitaly-goot:
Implemented shared key authentication scheme for BLOB service.
Authentication scheme described here:
https://docs.microsoft.com/en-us/rest/api/storageservices/authentication-for-the-azure-storage-services

To enforce signature verification one should start the emulator with '-a' option.

### DIFF
--- a/lib/core/env.js
+++ b/lib/core/env.js
@@ -20,6 +20,7 @@ class Environment {
         this.azuriteWorkspacePath = options.l || options.location || process.cwd();
         this.azuriteRootPath = path.join(__dirname, '../..');
         this.silent = options.s || options.silent;
+        this.accountAuth = options.a || options.accountAuth;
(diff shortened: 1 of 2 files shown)


edwin-huber:
Hi @vitaly-goot ,

many thanks for your patience.
Sorry, after fat fingering the review, I found that the issue is actually in our documentation of the API.

The some framework milddleware normalizes the headers for us.

In https://docs.microsoft.com/en-us/rest/api/storageservices/authenticate-with-shared-key, we specifically state that we should include 0 when evaluating content length, however this causes both the storage explorer and storage SDK tests to fail.

I'm figuring this out internally, and will update the pull request and then merge it.

edwin-huber:
Manually merged with changes.

Issue #26, opened 2018-05-26
Table Filters Returning No Results


bacon1986:
I'm using Azurite, Version 2.6.5 on Mac OSX

I perform the following request and I receive all results from the database:
```
GET /devstoreaccount1/Signposts HTTP/1.1
Host: 127.0.0.1:10002
Authorization: SharedKey devstoreaccount1:Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==
x-ms-date: 1527335502
Content-Type: application/json
x-ms-version: 2017-07-29
Cache-Control: no-cache
Postman-Token: 463d93a9-df19-49d1-8e59-eaf509432d41
```
```
{
    "value": [
        {
            "Timestamp": "2018-04-19T12:20:36.776Z",
            "title": "Benefits - GOV.UK",
            "description": "The best place to find government services and information. Simpler, clearer, faster",
            "url": "https://www.gov.uk/browse/benefits",
            "image": "/static/images/govuk.png",
            "categories": "Benefits",
            "PartitionKey": "Signposts",
            "RowKey": "574d5f400d3e106660158ac3"
        },
        {
            "Timestamp": "2018-04-19T12:20:36.791Z",
            "title": "Business & Self-Employment - GOV.UK",
            "description": "The best place to find government services and information. Simpler, clearer, faster",
            "url": "https://www.gov.uk/browse/business",
            "image": "/static/images/govuk.png",
            "categories": "Business",
            "PartitionKey": "Signposts",
            "RowKey": "574d5f400d3e106660158ac4"
        }
    ]
}
```

If I submit a request with a query (in this case on the PartitionKey) I get no results:

```
GET /devstoreaccount1/Signposts?$filter=PartitionKey%20eq%20%27Signposts%27 HTTP/1.1
Host: 127.0.0.1:10002
Authorization: SharedKey devstoreaccount1:Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==
x-ms-date: 1527335659
Content-Type: application/json
x-ms-version: 2017-07-29
Cache-Control: no-cache
Postman-Token: ae88e955-2304-409d-820f-60e5b0a7c6d3
```
```
{
    "value": []
}
```

[...]

PR #28, opened 2018-05-31
When PartitionKey and RowKey is passed in filter they are keept in Pa…

kalleep:
This is a fix for [Issue#26](https://github.com/Azure/Azurite/issues/26).

When PartitionKey and RowKey are passed in filter they are kept in pascal case, the db has them set to camel case


edwin-huber:
Good catch and thanks for the contribution!

PR #30, opened 2018-06-05
Fix an issue with CORS when a preflight request has an empty headers list

julienr:
Without this fix, azurite would fail because it was trying to call

  ```request.httpProps[N.ACCESS_CONTROL_REQUEST_HEADERS].toLowerCase()```

which will result in trying to call undefined.toLowerCase() if the preflight
request doesn't specify access-control-request-headers.

In addition, when validating a request against the CORS, the request would
fail to be validated if access-control-request-headers was empty.


edwin-huber:
Many thanks for your contribution!

PR #33, opened 2018-06-08
Correct readme references to the repo to the new location

esumerfd:
Not sure if this is appropriate but it looks like the issues are all on the new site now.


edwin-huber:
Hi @esumerfd  
Good catch!
Many  thanks for taking the time to correct the documentation.

PR #46, opened 2018-06-21
Entity generator now accepts optional partition and row key

Fanarito:
When the raw entity does not have a partition key or row key use the ones from the request params.

The C# azure storage sdk does not necessarily send the partition key and row key in the raw entity. Azure Storage apparently does not read it from the raw entity because code that works on it does not work on Azurite without these changes.

### DIFF
--- a/lib/core/table/TableStorageManager.js
+++ b/lib/core/table/TableStorageManager.js
@@ -190,7 +190,7 @@ class TableStorageManager {
 
     _createOrUpdateEntity(partitionKey, rowKey, tableName, rawEntity) {
         const coll = this.db.getCollection(tableName),
-            entity = EntityGenerator.generateEntity(rawEntity, tableName),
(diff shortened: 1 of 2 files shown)


edwin-huber:
Thanks for your contribution.

Based on the API definition, I'm not sure I understand what the problem is that you are fixing, as the partition and row ID are both required.  
 
[https://docs.microsoft.com/en-us/rest/api/storageservices/insert-entity](https://docs.microsoft.com/en-us/rest/api/storageservices/insert-entity)
```
When inserting an entity into a table, you must specify values for the PartitionKey and RowKey system properties. 
```
I was wondering if you have an example what works in Azure, but doesn't work against Azurite?
Thanks!

Fanarito:
We strip the row key and the partition key off the entity before sending it to azure table storage. That is the raw entity does not contain the keys but they are still send with the entity as query parameters.

edwin-huber:
OK, many thanks for the clarification.

Issue #53, opened 2018-06-26
Use Azure-Storage-Node in Tests


swissarmykirpan:
Rather than having to recreate the wheel with Http requests, why not use 
https://github.com/Azure/azure-storage-node ?

edwin-huber:
Hi @swissarmykirpan, many thanks for the suggestion.
That's a good idea.
Building out the TDD is on our list of things to do.
Right now we are using the .Net Storage SDK to validate the  functionality, but a Node version would make life easier.

edwin-huber:
Hi @pmanlukas , can you see if we can incorporate this and update the documentation accordingly?

pmanlukas:
Sure, i will investigate this week if we can switch to our SDK for that :)

swissarmykirpan:
@edwin-huber @pmanlukas I am happy to do this work if that helps :)

pmanlukas:
That would be a great suppport :) Then I would suggest you add the changes and make a pull request and we will review it than in time

edwin-huber:
@pdressel maybe you can connect with @swissarmykirpan , as you have been looking at this as well?

swissarmykirpan:
@edwin-huber @pmanlukas the problem with TDD is that effectively we would have the same tests that the node lib already does - or the c# client already does. Rather than re-implementing all of those tests, what if we forked some of the other client libs and then target a docker container with the tests?

pdressel:
@swissarmykirpan @edwin-huber @pmanlukas I agree, and azure-storage-node already comes with an exhaustive test suite for free. I was able to make it work by including node-azure-storage as a git submodule and writing a small shim to run their tests directly against Azurite via ´npm test´. This works with the existing pipeline without extra integration, and locking and updating the test code via git submodule is also easy to do. 

This could nicely complement PR#58, where we could write additional tests specific to Azurite not covered by the external tests. What do you think?

swissarmykirpan:
@pdressel where’s the branch/PR then? 😜

pmanlukas:
@pdressel is currently debugging the last issues with the storage tests. It is actually working to use his approach, but running the new storage tests showed some new issues in the code :) we keep you updated

pmanlukas:
@swissarmykirpan we just merged the first part of the implementation :)

edwin-huber:
We now have Azure Storage Node use in the basic testing, and are using azure-storage-node as our external integration tests.
Closing this issue.

Issue #52, opened 2018-06-25
Azure Table Query not working


swissarmykirpan:
I am doing a simple search on partitionkey and rowkey -> no data is being returned. If I remove all filters from the filter I get the data back.

I have tried this using `WindowsAzure.Storage` nuget package lib and also the Microsoft Azure Storage Explorer - both produce the same result

XiaoningLiu:
Hi @swagatmishra2007 

Can you help providing samples code used to query?

swagatmishra2007:
@XiaoningLiu sorry doesn't look related to anything i work on :)

XiaoningLiu:
@swagatmishra2007 Sorry.. 

@swissarmykirpan Can you help providing samples code used to query?

edwin-huber:
This "should" be solved in the latest release (current dev branch being promoted now)  
Please let us know in case we need to improve the test cases and associated table query code.

cswendrowski:
@edwin-huber Thanks for fixing this!

I went looking for the latest release in hopes of testing it out and noticed that both NPM and Dockerhub only have 2.6.5 (5 months old). It looks like there should have been a 2.6.6 deployed 2 days ago: https://travis-ci.org/Azure/Azurite/builds/421433522

Is the deployment having issues, or is there just a mechanism for promoting that I'm not seeing?

maciejgos:
Hi do you have any idea when v.2.6.6 will be available on npm? Thanks.

cswendrowski:
I am temporarily building and hosting the Docker images of this repository until deployments start coming out of this repo again: https://hub.docker.com/r/ironmoosedevelopment/azurite/

I can look into deploying the NPM version as well

XiaoningLiu:
Close this for long time noactive.

PR #58, opened 2018-07-16
refactor tests to use Azure-Storage-Node

swissarmykirpan:
https://github.com/Azure/Azurite/issues/53 - link to the issue

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.4.tgz",
       "integrity": "sha1-hiRnWMfdbSGmR0/whKR0DsBesh8=",
       "requires": {
-        "mime-types": "~2.1.16",
+        "mime-types": "2.1.17",
         "negotiator": "0.6.1"
       },
       "dependencies": {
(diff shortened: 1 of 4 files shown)


swissarmykirpan:
The test failure is because table storage not returning anything when being queried using partitionkey and rowkey - https://github.com/Azure/Azurite/issues/52

edwin-huber:
Interestingly, I can't reproduce the error that you and the travis build are reporting...

edwin-huber:
OK, found that the  "done" was not inside the callback from the  query. I'm looking into why we are not returning anything. Once fixed, I shall merge this.

edwin-huber:
hi @swissarmykirpan , the test should look something like this, but our table API needs quite a bit of work to return the correct values.
I'm working on getting it straightened out.
Thanks for your patience!

```javascript
    describe('GET Table Entity', () => {
        it('should retrieve one entity based on the PartitionKey and RowKey of testEntity 1, using OData Metadata None', (done) => {
            tableService.retrieveEntity(tableName, partitionKey, rowKey1, {payloadFormat:"application/json;odata=nometadata"},function (error, result, response) {
                console.log("RESUILT : # " , JSON.stringify(result));
                expect(result['PartitionKey']._).to.equal(partitionKey);
                expect(result['RowKey']._).to.equal(rowKey1);
                expect(result['description']._).to.equal(entityDescription1);
                expect(result['dueDate']._).to.equal(new Date(Date.UTC(2018, 12, 25)).toISOString().split('.')[0]+"Z" );
                done();
            });
        });
```

XiaoningLiu:
@edwin-huber 

@jiacfan has made the change and planned a release including this these days. : )

PR #62, opened 2018-07-30
Typescript Migration!

swissarmykirpan:
Got started with this, massive piece of work!

Added TSLINT and using Prettier also


lib/actions/blob/ListContainers.ts
```diff
@@ -53,32 +56,35 @@ class ListContainers {
       : (xmlContainerListModel.Marker = marker);
     // Fixme: We do not support markers yet
     delete xmlContainerListModel.NextMarker;
-    for (const container of containers) {
-      if (container.name === "$logs") {
+    for (const entity of entities) {
+      if (entity.container.name === "$logs") {
```
swissarmykirpan: attempting to break up a very convoluted object graph

lib/actions/queue/ClearMessages.ts
```diff
@@ -3,9 +3,7 @@ import AzuriteQueueResponse from "./../../model/queue/AzuriteQueueResponse";
 
 class ClearMessages {
   public process(request, res) {
-    const queue = QueueManager.getQueueAndMessage({
-      queueName: request.queueName
-    }).queue;
+    const queue = QueueManager.getQueueAndMessage(request.queueName).queue;
```
swissarmykirpan: getting ready for types by removing all passing in object literations `{}`

swissarmykirpan:
There is a lot of room for tests here

swissarmykirpan:
I have come up with a different strategy.

PR #64, opened 2018-08-06
fix resource release on SIGINT

danne931:
- Close method invocation was outside of execution context.  Updated process event handle callbacks to lambda expressions.
- Missing bluebird assignment in AzuriteTable


edwin-huber:
Hi @danne931 , 
many thanks for this contribution!
It would help us validate your changes if you could perhaps describe a test to validate this, or the reason for these changes?  
Are we causing node to leak resources without these changes?  
Would we not need to implement this logic for releasing resources in the code for all 3 services, not just the table API?  
Many thanks,  
Edwin

edwin-huber:
Created new PR to resolve this.

PR #68, opened 2018-08-22
Testing using Azure-Storage-Node submodule

edwin-huber:
These changes implemented by Philipp allow us to use azure-storage-node test via a shim as an integration test provider.
It allows us to have a reliable set of tests from the Azure node SDK to direct our development work, as the  original Azurite code base was written without very much testing.  

**This does not preclude the need to provide good tests with new code that we are writing!**  

Documentation on using these tests is provided in a separate pull request.


swissarmykirpan:
this is nice!

Issue #27, opened 2018-05-28
Could we format logs better?


longtimeago:
Currently (working with 0.9.7), if I open Docker container logs in a text reader I see something like
```
[0mGET /account/customer?restype=container&comp=list [32m200 [0m19.399 ms - 255[0m
```
Escape chars are so annoying :(
Could we turn them off?

Also, would be really helpful to print timestamp in some way, e.g. `HH:mm:ss.SSS`

Thanks a lot!

longtimeago:
Maybe, we could have an environment variable to optionally switch log coloring off

edwin-huber:
Many thanks for your suggestion.

What exactly are you looking to get out of the logging / how are you using it?
This will help me define some requirements and scope to improve it.
Logging is turned off with the --silent switch.
Ascii greeting is in lib/core/cli.js

logging is controlled by 'morgan' in AzuriteBlob.js, AzuriteQueue.js, AzuriteTable.js 
like this:
```javascript
const app = express();
                if (!env.silent) {
                    app.use(morgan('dev'));
                }
```

longtimeago:
@edwin-huber Thanks for the hint :)

I've mixed 2 issues in one (sorry for that)

1. I'd like to have a way to turn off log coloring via (for example) special env variable. So, when opening logs in ordinary text editor instead of 
```
�[0mGET /account/customer?restype=container&comp=list �[32m200 �[0m19.399 ms - 255�[0m
```
I would see 
```
GET /account/customer?restype=container&comp=list 32m19.399 ms 
```
Usually, we forward our logs to logstash, where this coloring not needed as well :)

2. In each log entry instead of a relative time from container start would be very nice to have absolute timestamp. Otherwise, it's really tricky to match log entries in different log files

edwin-huber:
Thanks.
I'll mark this as a good first issue and hopefully somebody can get around to  looking at it soon.

We use morgan middleware for the logging, which is documented here:
https://github.com/expressjs/morgan

to give you the date and unformatted version like you are asking for, one would need to usesomething like:

```javascript
if (!env.silent) {
    app.use(morgan(':date[iso] :method :url :status :res[content-length] - :response-time ms'));
} 
```

pmanlukas:
I will look into the logging

pmanlukas:
I will start implementing this change in this week

PureKrome:
Was there any resolution to this issue?

[...]

PR #79, opened 2018-08-24
Adjust logging to have Standard Apache common log output.

pmanlukas:
I worked on Issue #27 and changed from dev to apache common logging format. Now color is gone and logs shouldn't cause problems.


pmanlukas:
I see I just thought coloring would be the issue. Will adjust it

edwin-huber:
Thanks 😄

longtimeago:
@pmanlukas would we get timestamp for each log entry with this change?

Issue #20, opened 2018-04-27
Table operation errors


craxal:
I’ve been playing with Azurite some, and there appears there may be a problem with it’s tables implementation. Regardless of platform, I get errors when trying to add entities, delete entities, and delete tables. Some details:
- We use [azure-storage-node](https://github.com/Azure/azure-storage-node) for sending requests.
- We expect a certain result object when adding entities succeeds. However, we are getting an empty result object. This result object is supposed to provide information about the new entity, such as an etag.
- When attempting to delete entities or tables, we get errors like the following (may be related to #17, #19):
```
POST /devstoreaccount1/test01 204 2.216 ms - -
GET /devstoreaccount1/test01?%24top=1000 200 1.560 ms - 575
GET /devstoreaccount1/test01?%24top=1000 200 0.747 ms - 575
POST /devstoreaccount1/$batch 500 1.124 ms - 45
**PANIC** Something unexpected happened! Table Storage Emulator may be in an inconsistent state!
TypeError: Cannot read property 'tableName' of undefined
    at BbPromise.try (C:\Users\cralvord\AppData\Roaming\npm\node_modules\azurite\lib\middleware\table\validation.js:24:48)
    at tryCatcher (C:\Users\cralvord\AppData\Roaming\npm\node_modules\azurite\node_modules\bluebird\js\release\util.js:16:23)
    at Function.Promise.attempt.Promise.try (C:\Users\cralvord\AppData\Roaming\npm\node_modules\azurite\node_modules\bluebird\js\release\method.js:39:29)
    at module.exports (C:\Users\cralvord\AppData\Roaming\npm\node_modules\azurite\lib\middleware\table\validation.js:18:18)
    at Layer.handle [as handle_request] (C:\Users\cralvord\AppData\Roaming\npm\node_modules\azurite\node_modules\express\lib\router\layer.js:95:5)
    at trim_prefix (C:\Users\cralvord\AppData\Roaming\npm\node_modules\azurite\node_modules\express\lib\router\index.js:317:13)
    at C:\Users\cralvord\AppData\Roaming\npm\node_modules\azurite\node_modules\express\lib\router\index.js:284:7
    at Function.process_params (C:\Users\cralvord\AppData\Roaming\npm\node_modules\azurite\node_modules\express\lib\router\index.js:335:12)
    at next (C:\Users\cralvord\AppData\Roaming\npm\node_modules\azurite\node_modules\express\lib\router\index.js:275:10)
[...]

Issue #19, opened 2018-04-25
*PANIC* when looking for non-existent row


tomasaschan:
There seems to be an inconsistency in how this emulator handles non-existent rows compared to how "real" Azure Table Storage handles them. I discovered this while using [Streamstone](https://github.com/yevhen/Streamstone), which builds event streaming capabilities on top of Azure Table Storage, and while I haven't been able to discern exactly what operations they're doing on the table storage, I have been able to build a minimal example that reproduces the problem.

To reproduce, make sure you have the `dotnet` CLI installed, and run

```
mkdir repro && cd repro
dotnet new console
dotnet add package Streamstone
```

Then replace the contents of `Program.cs` with this:

```
using System;
using System.Threading.Tasks;
using Microsoft.WindowsAzure.Storage;
using Streamstone;

namespace AzuritePanicOnStreamstoneTryOpenStream
{
    class Program
    {
        public static async Task Main(string[] args)
        {
            // Replace with CloudStorageAccount.Parse("your connection string"); for using a real Azure Storage Account
            var account = CloudStorageAccount.DevelopmentStorageAccount;

            var table = account
                .CreateCloudTableClient()
                .GetTableReference("Events");
            await table.CreateIfNotExistsAsync();

            var partition = new Partition(table, "foo");

            await Stream.TryOpenAsync(partition);
        }
    }
}
```

Finally, run `dotnet run` to see the program fail with

```
Unhandled Exception: Microsoft.WindowsAzure.Storage.StorageException: Unexpected response code, Expected:OK or NotFound, Received:InternalServerError
   at Microsoft.WindowsAzure.Storage.Core.Executor.Executor.ExecuteAsyncInternal[T](RESTCommand`1 cmd, IRetryPolicy policy, OperationContext operationContext, CancellationToken token) in C:\Program Files (x86)\Jenkins\workspace\release_dotnet_master\Lib\WindowsRuntime\Core\Executor\Executor.cs:line 315
   at Streamstone.Stream.OpenStreamOperation.ExecuteAsync()
   at AzuritePanicOnStreamstoneTryOpenStream.Program.Main(String[] args) in C:\Work\OSS\Throwaway repros\AzuritePanicOnStreamstoneTryOpenStream\Program.cs:line 17
   at AzuritePanicOnStreamstoneTryOpenStream.Program.<Main>(String[] args)
```

If I look at the logs for my docker container running Azurite, I see

```
[...]

PR #80, opened 2018-08-27
Fixes for table queries not returning 404 for missing entities

edwin-huber:
Added test case for missing entities, when using direct reference or query of objects in Table.

Added corresponding logic to return 404

Should address Issues #20 and #19


edwin-huber:
Should also resolve #26

test/01_Basic_table_Tests.js
```diff
@@ -109,7 +110,7 @@ describe('Table HTTP Api tests', () => {
         }
 
         it('should retrieve all Entities', (done) => {
-            const query = new azureStorage.TableQuery();
+            let query = new azureStorage.TableQuery();
```
XiaoningLiu: why change `const` to `let`? For non changed variables, we should use `const` instead of `let`.

PR #90, opened 2018-09-05
Refactored error handling

pdressel:
Introduces a new AzuriteErrorRespone, subclassed from AzuriteResponse, that can correctly package and send errors back to the client.

Also sets the x-ms-error-code header and adjusts the API version accordingly to improve compatibility with newer clients.

Finally, the AzuriteResponse class gains a send method, which will apply all headers, set status code, calculate content-length and send the data over the wire via an express `res` object. Over time, all Azurite actions can migrate to use this pattern, instead of manually applying all properties from the AzuriteResponse.

### DIFF
--- a/lib/core/AzuriteError.js
+++ b/lib/core/AzuriteError.js
@@ -7,12 +7,13 @@ function generateErrorMessage(errorCode, userMessage) {
 }
 
 class AzuriteError extends Error {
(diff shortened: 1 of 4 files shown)


edwin-huber:
LGTM -> Merging 👍

PR #91, opened 2018-09-05
Fix issue 70

pdressel:
Fixes issue #70 by simulating a storage container being in the deleting state for 5 seconds after the delete request.


edwin-huber:
Looks like ErrorCodes.js and  lib/core/blob/StorageManager.js needs fixing up after merge...

pdressel:
Fixed, sorry i missed that. After merging latest dev, now the test is broken again 😅
Another fix incoming

PR #99, opened 2018-09-28
SAS Logic

edwin-huber:
The changes support the .Net samples as requested by Sercan for SAS auth.

I still need to implement code to handle the following 2 calls from the samples:
```csharp
await ConfigureBlobAnalyticsAsync(blobClient);
await GetServiceStatsForSecondaryAsync(blobClient);
```
We should finish that and then release again, as there seem to be quite a few issues opened which are all solved in the dev branch. 😇

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -27,6 +27,22 @@
             "type": "node",
             "request": "launch",
             "protocol": "auto",
+            "name": "Launch Azurite with AuthEnabled",
+            "program": "${workspaceRoot}/bin/azurite",
+            "cwd": "${workspaceRoot}",
+            "args": [
+                "-l",
+                "${workspaceRoot}/azurite-testdrive",
+                "-a"
+            ],
+            "skipFiles": [
+                "node_modules/**/*.js"
+            ]
+        },
+        {
+            "type": "node",
+            "request": "launch",
+            "protocol": "auto",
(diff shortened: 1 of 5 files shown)


.vscode/launch.json
```diff
@@ -47,7 +63,8 @@
             "cwd": "${workspaceRoot}",
             "args": [
                 "-l",
-                "${workspaceRoot}/azurite-testdrive"
+                "${workspaceRoot}/azurite-testdrive",
+                "-a"
```
pdressel: Is auth intentionally on here?

lib/middleware/blob/authentication.js
```diff
@@ -51,13 +35,41 @@ module.exports = (req, res, next) => {
...
+      request.auth.usingAccountSAS = true;
+      request.auth.sasValid = true;
+    } else if(sigService === request.query.sig){
+      request.auth.usingServiceSAS = true;
+      request.auth.sasValid = true;
+    }
+    
+    if (env.accountAuth && req.method !== "HEAD" && request.auth.sasValid === false) {
+      if (req.headers.authorization === undefined)
+         throw new AError(ErrorCodes.AuthenticationFailed);
```
pdressel: I would advocate for putting this in an explicit scope block just to be sure this does not get messed up.

PR #102, opened 2018-10-11
storage account and storage account key

SkinyMonkey:
storage account and storage account key can now be specified through environment variables.

All tests pass and hand ones were made for upload/download of blobs.


edwin-huber:
Thanks for the submission! 😄 
It all looks good to me, but while there's no test cases associated specifically with this submission I tested with authEnabled and all passes.
I'll merge with the SAS Updates
@XiaoningLiu  @pdressel  any comments?

XiaoningLiu:
Azurite V3 supports account name/key customization, closed this since long time no update.

PR #105, opened 2018-10-19
Updated paths so tests don't rely on cloning directory

fearphage:
The tests are currently set up to require you to clone this repo into a folder named `Azurite`. If you do not, the tests will not run. This removes that requirement by making the test paths relative.


edwin-huber:
Thanks for the contribution and catching this! :)

PR #31, opened 2018-06-08
Fixed result when requesting single entity with rowkey and partitionkey

kalleep:
When using this with azure-storage-node and calling the api with retrieveEntity the wrong format is returned, it handles the response as it does when querying several entities

Return value when requesting one entity from dev

```json
{
	"value": {
		"_": [{
			"Timestamp": "2018-06-08T10:09:31.010Z",
			"Timestamp@odata.type": "Edm.DateTime",
			"type": 1,
			"PartitionKey": "asdasd123",
			"RowKey": "a6c09770-6a69-11e8-ac4a-ddbdcad9d221"
		}]
	},
	".metadata": {
		"metadata": "http://127.0.0.1:10002/devstoreaccount1/$metadata#revokelist",
		"etag": "W/\"fe-/JfNhqfuZdZakFgYTVFcUMDIGdc\""
	}
}
```
Return value when requesting one entity from azure table
```json
{
	"PartitionKey": {
		"$": "Edm.String",
		"_": "asdasd"
	},
	"RowKey": {
		"$": "Edm.String",
		"_": "839408a0-6a68-11e8-a3a3-fd722a41ad6b"
	},
	"Timestamp": {
		"$": "Edm.DateTime",
		"_": "2018-06-07T15:38:11.801Z"
	},
	"type": {
		"_": 1
	},
	".metadata": {
		"metadata": "https://{url}/$metadata#revokelist/@Element",
		"etag": "W/\"datetime'2018-06-07T15%3A38%3A11.8019139Z'\""
	}
}
```

Return value of one entity with fix
```json
{
	"PartitionKey": {
		"$": "Edm.String",
		"_": "asdasd123"
	},
	"RowKey": {
		"$": "Edm.String",
		"_": "a6c09770-6a69-11e8-ac4a-ddbdcad9d221"
	},
	"Timestamp": {
		"$": "Edm.DateTime",
		"_": "2018-06-08T10:09:31.010Z"
	},
	"type": {
		"_": 1
	},
	".metadata": {
		"metadata": "http://127.0.0.1:10002/devstoreaccount1/$metadata#revokelist",
		"etag": "W/\"f2-OvSglgZEq3Cvo1EjrlNQtBG4j6o\""
	}
}
```
RowKey and PartitionKey will only be set in header when calling with retrieveEntity

PR #106, opened 2018-10-19
Fix query single item incorrect behavior

alexswan10k:
Appears to be from commit 9413bd7af4da7ffe24457180541ad1c6d2da8d10 for the 2.6.7 release, special behavior was introduced for single results. This breaks some clients as they are not expecting a structural change in the resulting payload, which appears to be inconsistent with the Microsoft official one.

I cannot find out exactly why this change was made as the history doesn't appear to be traceable beyond this, but I appreciate there might be a legitimate reason for this.

May be connected to #31.

I have a specific fake 5 F# test example I have attached below which will no longer explode after the change:

```
open System
open Microsoft.WindowsAzure.Storage
#load "./.fake/T1.fsx/intellisense.fsx"

#r "paket:
nuget FSharp.Azure.Storage //"

open FSharp.Azure.Storage.Table

type Thing = {
    [<PartitionKey>] PK: string
    [<RowKey>] Id: string
    string1: Option<string>
    string2: Option<string>
    string3: Option<string>
}

open Microsoft.WindowsAzure.Storage
open Microsoft.WindowsAzure.Storage.Table

let account = CloudStorageAccount.Parse "UseDevelopmentStorage=true;"
let tableClient = account.CreateCloudTableClient()
tableClient.GetTableReference("Thing").CreateIfNotExistsAsync().Result

let fromThingTable thing = fromTable tableClient "Thing" thing
let inThingTable thing = inTable tableClient "Thing" thing

let thing1 = {PK = "ONE"; Id = "THREE"; string1 = Some "hello world"; string2 = Some "test"; string3 = None}
printfn "before"
let result = thing1 |> InsertOrReplace |> inThingTable
let q2 = 
    Query.all<Thing>
    |> Query.where <@ fun g sf -> g.Id = "THREE" @>
    |> Query.take 1 
    |> fromThingTable 

printfn "%A" q2
```

### DIFF
--- a/lib/actions/table/QueryEntities.js
+++ b/lib/actions/table/QueryEntities.js
@@ -37,13 +37,7 @@ class QueryEntities {
         response["RowKey"] = item.rowKey;
         response = Object.assign({}, response, item.attribs(accept));
       }
-    } else if (payload.length === 1) {
-      response["PartitionKey"] = payload[0].partitionKey;
-      response["RowKey"] = payload[0].rowKey;
-      for (const attrib in payload[0].attribs(accept)) {
-        response[attrib] = payload[0]._.attribs[attrib];
-      }
-    } else if (payload.length > 1) {
+    } else {
       response.value = [];
       let i = 0;
       for (const item of payload) {


seniorquico:
I also haven't observed this different "single result" model while capturing traffic between Azure Storage and the Azure Storage Explorer using Fiddler. The results always seem to come back as an object in an array.

However, of additional note, this PR fixes another issue I encountered today. Namely, the Azure Storage Explorer and one of our C# clients appear expect a result with an empty array when the table is empty. I have observed that Azure Storage returns results this way:

```
HTTP/1.1 200 OK
Cache-Control: no-cache
Content-Type: application/json;odata=minimalmetadata;streaming=true;charset=utf-8
Server: Windows-Azure-Table/1.0 Microsoft-HTTPAPI/2.0
x-ms-request-id: 00000000-0000-0000-0000-000000000000
x-ms-version: 2018-03-28
X-Content-Type-Options: nosniff
Date: Thu, 22 Nov 2018 09:43:54 GMT
Connection: close
Content-Length: 101

{"odata.metadata":"https://storage.table.core.windows.net/$metadata#Example","value":[]}
```

[This addresses](https://github.com/Azure/Azurite/pull/106/files#diff-afa4e12f449014de3f956525bc42a642L46) this second issue where the `value` is omitted when the table is empty.

edwin-huber:
Many thanks for the contribution! :smile:
The changes have now been imported.
They match the API description here: 
https://docs.microsoft.com/en-us/rest/api/storageservices/query-entities

alexswan10k:
Glad I could help :)

PR #118, opened 2018-11-22
Fixed errors when querying tables with a $filter argument

### LINKED ISSUE #117: Failing to open a new Azurite table in Storage Explorer
I have Node.js 10.13.0 installed via nvm. I started Azurite 2.7.0 by running the following:

```sh
$ node --version
v10.13.0
$ mkdir empty
$ cd empty
$ npx azurite
...
```

Azurite started normally. I then used Storage Explorer 1.5.0 to create a new table named "Example". The operation completed successfully, and I saw the "Example" table in the navigation tree. However, selecting the table to open it resulted in the following error:

```
<?xml version="1.0" encoding="utf-8"?><Error><Code>EntityNotFound</Code><Message>The specified entity does not exist.</Message></Error>
```

I found the following in the console log (no other errors or anything that looked too relevent):

```
POST /devstoreaccount1/Tables 204 2.419 ms - -
GET /devstoreaccount1/Example?%24top=1000 404 0.866 ms - 135
```

This is a simple reproduction of a more general issue we've encountered creating and reading tables in Azurite. Working with tables no longer appears to be working from our .NET Core application.

seniorquico:
Fixes #117 and #120. For #117, I tested by manually querying using the following two URL patterns:

```
/TableName(PartitionKey='pk',RowKey='rk')
/TableName?$select=&$filter=PartitionKey eq 'pk' and RowKey eq 'rk'&$top=1000
```

This is my first time looking at both Azurite and the Azure Storage API. This is my best guess as to how the Azure Storage API behaves based purely on Fiddler traffic captures of the Azure Storage Explorer. Guidance is appreciated if there's a better method to debugging and fixing an issue like this.

I haven't yet looked at the Azurite test suite, but I'm guessing the second URL pattern isn't covered given it isn't working on the latest release. I'll try to amend this PR to include an additional test.

EDIT: I updated this PR to include a fix for #120 as the unit test behavior is also dependent on this specific response. The fix for #120 is to simply escape backticks.

### DIFF
--- a/lib/actions/table/QueryEntities.js
+++ b/lib/actions/table/QueryEntities.js
@@ -43,7 +43,7 @@ class QueryEntities {
       for (const attrib in payload[0].attribs(accept)) {
         response[attrib] = payload[0]._.attribs[attrib];
       }
-    } else if (payload.length > 1) {
+    } else {
       response.value = [];
       let i = 0;
       for (const item of payload) {
--- a/lib/core/table/TableStorageManager.js
+++ b/lib/core/table/TableStorageManager.js
@@ -145,7 +145,7 @@ class TableStorageManager {
         .data();
     }
 
-    if (find.length == 0) {
+    if (request.partitionKey && request.rowKey && find.length == 0) {
       throw new AError(ErrorCodes.EntityNotFound);
     }
 
--- a/lib/model/table/AzuriteTableRequest.js
+++ b/lib/model/table/AzuriteTableRequest.js
@@ -99,6 +99,8 @@ class AzuriteTableRequest {
       // ignoring these query keywords since we compare simply on a string-level
       .replace(/\bdatetime\b/g, "")
       .replace(/\bguid\b/g, "")
+      // Escape a single backtick to prevent interpreting the start of a template literal.
+      .replace(/`/g, '\\`')
       // A simple quotation mark is escaped with another one (i.e. '').
       // Since we will evaluate this string we replace simple quotation marks indictaing strings with template quotation marks
       .replace(/''/g, "@")
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
     "externaltests": "env-cmd ./externaltests/.env mocha externaltests",
     "start": "node bin/azurite -l azurite-testdrive",
     "blob": "node bin/blob -l azurite-testdrive",
-    "queue": "node bin/queue",
-    "table": "node bin/table",
+    "queue": "node bin/queue -l azurite-testdrive",
+    "table": "node bin/table -l azurite-testdrive",
     "clean": "rimraf azurite-testdrive *.nupkg blob.exe queue.exe table.exe",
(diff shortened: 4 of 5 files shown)


seniorquico:
Bummer, I accidentally ran the tests on a local branch that also included the changes in #119. It appears this won't work as I'm expecting without also including that change.

edwin-huber:
hi, thanks for the contribution.
There might be an easier way to validate the responses:
If you look at the chai examples from test/02_00_Basic_blob_Tests.js you can see how we validate requests against the REST Api using an HTTP request based pattern, rather than the Azure Storage Node SDK which we moved to in later tests.

I'm having a look at these changes, however Azure Storage should return 404 for EntityNotFound... can you maybe provide a separate test case which we can use to validate the behavior you describe vs that which we have?

When I run the test using Azure Storage Node against Azure, I get a 404 not found, rather than an empty response.

seniorquico:
Thanks, @edwin-huber. I just re-ran the test using Azure CLI against Azure Storage with Fiddler in the middle capturing traffic, and I can confirm that I see the behavior described in the linked issue. To be clear, I _do_ get a 404 Not Found when using the following URL syntax:

```
/Contact(PartitionKey='Smith',RowKey='John')
```

However, I _do not_ get a 404 Not Found when using the following URL syntax (run through urldecode just to make it easier to read):

```
/Contact()?$filter=PartitionKey eq 'Smith' and RowKey eq 'John'
```

Currently, Azurite always answers with the former (a 404 Not Found) instead of two different responses (404 Not Found and 200 OK) for each of these two URLs.

I started the test by creating a new Azure Storage account (general purpose v1, standard performance, RA-GRS in East+West US). What follows is the test script I used along with the responses printed to stdout. I used Azure CLI 2.0.50 on both Windows and Linux (Ubuntu). I will post the Fiddler capture to a gist a little later this afternoon.

```sh
$ export AZURE_STORAGE_CONNECTION_STRING=xxx
$ az storage entity show --table-name Contact --partition-key Smith --row-key John
{"odata.error":{"code":"TableNotFound","message":{"lang":"en-US","value":"The table specified does not exist.\nRequestId:00000000-0000-0000-0000-000000000000\nTime:2018-11-27T20:52:39.7304095Z"}}}
$ az storage table create --name Contact
{"created":true}
$ az storage entity show --table-name Contact --partition-key Smith --row-key John
{"odata.error":{"code":"ResourceNotFound","message":{"lang":"en-US","value":"The specified resource does not exist.\nRequestId:00000000-0000-0000-0000-000000000000\nTime:2018-11-27T20:53:10.1535873Z"}}}
$ az storage entity query --table-name Contact --filter "PartitionKey eq 'Smith' and RowKey eq 'John'"
{"items":[],"nextMarker":{}}
```

Describing it with the Azure Storage Node SDK methods, this should be calling `retrieveEntity` for the former whereas the latter should be using a `TableQuery` with `where`.

I found this (and a couple of other issues) trying to run Orleans storage/membership/streams providers against Azurite. The Orleans providers are built using the Azure Storage SDK for .NET (the `Azure.Storage` NuGet package) using storage version 2018-03-28. I can confirm that the Orleans providers fail with Azurite as-is, but the Orleans providers succeed with this change (and others).

edwin-huber:
Many thanks for the contribution! 😄 
Now merged into the dev branch.

PR #119, opened 2018-11-22
Fixes incorrect handling of table management URLs

seniorquico:
While debugging, I noticed that some table operations were being incorrectly handled by both the `TableRoute` and `EntityRoute` handlers. For example, I encountered problems deleting tables using Azure Storage Explorer.

```
DELETE /Tables('Example')
```

On the first pass, the `TableRoute` handler correctly parsed the request as a `DELETE_TABLE` operation with a `AzuriteTableRequest.tableName` equal to "Example". However, the call to express's `next()` caused the request to flow to `EntityRoute`. On this second pass, the `EntityRoute` handler incorrectly parsed the request as a `DELETE_ENTITY` operation with a `AzuriteTableRequest.tableName` equal to "Tables". It was this second, incorrect request that was actually processed and resulted in an error (because a table named "Tables" didn't exist).

The issue stems from the overlap of RegExp matching of these two different URL patterns. (I noticed that the Blob and Queue endpoints don't have this same issue.) I'm was able to quickly fix and test this solution with Azure Storage Explorer and a C# client. I'm not well versed in express, feedback welcome if there's a better approach!

### DIFF
--- a/lib/routes/table/EntityRoute.js
+++ b/lib/routes/table/EntityRoute.js
@@ -15,7 +15,7 @@ const env = require("./../../core/env"),
 module.exports = (app) => {
   app
     .route(
-      new RegExp(`\/${env.emulatedStorageAccountName}\/(?![Tt][Aa][Bb][Ll][Ee][Ss])([A-Za-z0-9]+)(.*)`)
+      new RegExp(`\/${env.emulatedStorageAccountName}\/([A-Za-z0-9]+)(.*)`)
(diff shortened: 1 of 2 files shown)


seniorquico:
Ah, I missed before how the other route methods in `EntityRouter` check to see if a `req.azuriteOperation` has already been defined. This check was missing on the `delete` method. I updated the PR to continue with that approach.

PR #124, opened 2018-11-30
Made etag handling same as in microsoft

### LINKED ISSUE #95: Exception parsing ETag of Azurite response
We are using the latest version of the Docker Azurite. We can access it through the Azure Explorer but...

Azurite is returning a 200 response with: 
{ETag: W/"75-mAoh3qooM6Mzsi98Dku2rmIXONw"}

As opposed to the same response coming from the real Azure with the format: 
{ETag: W/"datetime'2018-09-07T14%3A42%3A56.757613Z'"}

Is this a know issue, a new issue or are we doing something wrong?

michaelkruglos:
Fixes #95

### DIFF
--- a/lib/core/utils.js
+++ b/lib/core/utils.js
@@ -2,12 +2,6 @@
 
 "use strict";
 
-const crypto = require("crypto");
-
 exports.computeEtag = (templateString) => {
-  return crypto
-    .createHash("sha1")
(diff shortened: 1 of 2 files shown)


edwin-huber:
Many thanks for the contribution! 😄 
Changes now merged into dev

PR #130, opened 2018-12-03
Fix query entities for single result in a collection

### LINKED ISSUE #129: The Query Entities operation returns the list of entities in a table
Microsoft Azure Storage REST API states https://docs.microsoft.com/en-us/rest/api/storageservices/query-entities#response-body:

> The Query Entities operation returns the list of entities in a table as an OData entity set, either in a JSON format or an Atom feed, depending on the Accept header of the request.

This behavior is present in Azure Storage Emulator:

```
HTTP/1.1 200 OK
Cache-Control: no-cache
Transfer-Encoding: chunked
Content-Type: application/json;odata=minimalmetadata;streaming=true;charset=utf-8
Server: Windows-Azure-Table/1.0 Microsoft-HTTPAPI/2.0
x-ms-request-id: 8eefc82a-7169-49d9-9529-4a9c5f5e83c0
x-ms-version: 2018-03-28
X-Content-Type-Options: nosniff
Date: Mon, 03 Dec 2018 13:45:21 GMT

109
{"odata.metadata":"http://127.0.0.1:10002/devstoreaccount1/$metadata#test","value":[{"odata.etag":"W/\"datetime'2018-12-03T13%3A43%3A09.65Z'\"","PartitionKey":"PartitionKeyForEntity","RowKey":"RowKeyForEntity","Timestamp":"2018-12-03T13:43:09.65Z","Data":"Value"}]}
0
```

But Azurite returns an entity directly, without wrapping into value collection.

gpatryk:
Fixes #129

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -26,6 +26,25 @@
         {
             "type": "node",
             "request": "launch",
+            "env": {
+                "AZURITE_LOCATION": "azurite-testdrive"
+            },
+            "name": "Mocha Current File",
+            "program": "${workspaceRoot}/node_modules/mocha/bin/_mocha",
+            "args": [
+                "-u",
+                "tdd",
+                "--timeout",
+                "999999",
+                "--colors",
+                "${file}"
+            ],
+            "internalConsoleOptions": "openOnSessionStart",
+            "protocol": "inspector"
+        },
+        {
+            "type": "node",
+            "request": "launch",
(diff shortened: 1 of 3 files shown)


seniorquico:
Appears to be the same as #106. This PR includes a unit test, though. 👍

gpatryk:
You are right @seniorquico. I must have overlooked this somehow.

PR #134, opened 2018-12-07
Correct OData Metadata for single entity

### LINKED ISSUE #133: Incorrect OData Metadata for single entity response
When metadata are requested Azurite always returns odata.metadata = `http://127.0.0.1:10002/devstoreaccount1/$metadata#TableName`. For single entity result is should be `http://127.0.0.1:10002/devstoreaccount1/$metadata#TableName/@Entity`.

gpatryk:
Fixes #133 

I have also removed hardcoded development storage connection string in Table service unit tests.


edwin-huber:
Merged

PR #139, opened 2018-12-12
Send response using AzuriteResponse object for getBlockList

### LINKED ISSUE #137: getBlockList crashes azure-storage js sdk
When calling getBlockList Azurite fails to set the Content-Type header. It is set as `text/html` instead of `application/xml` per Azure. This stems from the response being sent using the ExpressJS response object instead of the AzuriteResponse object so headers set with `response.addHttpProperty` are never applied to the ExpressJS response object. This happens in [lib/actions/getBlockList.js](https://github.com/Azure/Azurite/blob/f3dccf71dae73d1af8bc355e39c215515a322d4f/lib/actions/blob/GetBlockList.js#L22)

tmacro:
Fixes #137


XiaoningLiu:
Closed this since long time no update.

PR #140, opened 2018-12-18
fixing headers for azure java client library

stevenleadbeater:
"x-ms-blob-content-md5":"undefined" is returned in the headers for get requests on blobs with this line commented out. When using the com.microsoft.rest.v2:client-runtime:2.0.0 maven dependency (which is used by the com.microsoft.azure:azure-storage-blob:10.1.0 library to make rest calls), `com.microsoft.rest.v2.protocol.HttpResponseDecoder` will fail at line 259 while trying to deserialise the value "undefined" in to the `byte[]` property `blobContentMD5` on `com.microsoft.azure.storage.blob.models.BlobDownloadHeaders`


lib/middleware/blob/validation.js
```diff
@@ -67,6 +67,7 @@ module.exports = (req, res, next) => {
     // Refactor me: Move this to bin/azurite (exception needs to carry res object), and handle entire exception handling there
   }).catch((e) => {
     res.status(e.statusCode || 500).send(e.message);
+    console.log(e.message);
```
XiaoningLiu: Remove console.log

lib/actions/blob/GetBlob.js
```diff
@@ -42,26 +44,18 @@ class GetBlob {
       );
       if (request.auth) response.sasOverrideHeaders(request.query);
 
-      // If x-ms-range-get-content-md5 is specified together with the range attribute we load the entire data range into memory
+      // If there is no precomputed md5 hash we load the entire data range into memory
       // in order to compute the MD5 hash of this chunk. We cannot use piping in this case since we cannot modify the HTTP headers
       // anymore once the response stream has started to get delivered.
       // Otherwise we just pipe the result through to the client which is more performant.
-      if (range && request.httpProps[N.RANGE_GET_CONTENT_MD5]) {
```
XiaoningLiu: Why remove range content md5 related code?

stevenleadbeater: because the way this is implemented, if you don't specify the MD5 in the GET or the original POST (which is how the java rx based client library works) then you don't get any MD5 at all and an "undefined" is returned in the headers. This isn't right, it works, really we need to change the response headers to omit the md5 header if it is undefined, however, I haven't been able to manage that. Do you know where in the code I could do it?

XiaoningLiu:
Close this for legacy Azurite V2. Please use Azurite V3.

PR #148, opened 2019-01-21
Fix close() to return a promise

### LINKED ISSUE #144: close() method of AzuriteBlob, AzuriteQueue and AzuriteTable won't wait for resolution
Currently, the `close()` methods for `AzuriteBlob`, `AzuriteQueue` and `AzuriteTable`, although they return a `Promise`, the code within the promise won't await for the methods called within it:

https://github.com/Azure/Azurite/blob/f3dccf71dae73d1af8bc355e39c215515a322d4f/lib/AzuriteBlob.js#L86-L92

https://github.com/Azure/Azurite/blob/f3dccf71dae73d1af8bc355e39c215515a322d4f/lib/AzuriteQueue.js#L58-L62

https://github.com/Azure/Azurite/blob/f3dccf71dae73d1af8bc355e39c215515a322d4f/lib/AzuriteTable.js#L65-L71

Issues:

- `this.server.close()` expects a callback that will be called on completion. It should be promisified.
- Both `storageManager.flush()` and `tableStorageManager.flush()` return the promise of `db.saveDatabaseAsync()`, but is not awaited for.
- Both `storageManager.close()` and `tableStorageManager.close()` return the results of `db.close()`, but it should really return the results of `db.closeAsync()`.

Finally, where more than 1 method is called, they should be wrapped with `BbPromise.all()`.

If I'm not missing anything, then I could prepare a PR for you.

poveden:
Fixes #144

### DIFF
--- a/lib/AzuriteBlob.js
+++ b/lib/AzuriteBlob.js
@@ -84,11 +84,11 @@ class AzuriteBlob {
   }
 
   close() {
-    return BbPromise.try(() => {
-      this.server.close();
-      storageManager.flush();
-      return storageManager.close();
-    });
+    return new BbPromise((resolve, reject) => {
+      this.server.close(err => err ? reject(err) : resolve());
+    })
+    .then(() => storageManager.flush())
+    .then(() => storageManager.close());
   }
 }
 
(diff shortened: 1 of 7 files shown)


edwin-huber:
Many thanks for the contribution.
The table tests no longer completed, as the promise for table storage was taking overly long to resolve on completion of the tests.
I modified the promise to return a promise all, and this now solved the test response time.
These changes were merged into dev. :smile:

PR #154, opened 2019-02-17
New arch doc update

edwin-huber:
Update to notify about new architecture.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ The plan is to implement blob storage, followed by table storage, then queues.  
 
 Contributors and maintainers can then focus on emulation and simulation of storage behaviour, rather than reverse engineering our APIs and API documenation.  
 
+The blob API work should be completed in the next couple of months, with the rest following on close behind.
+
 # Installation and Usage
 ## NPM
 In order to run Azurite you need [Node.js](https://nodejs.org/) >= 6.0 installed on your system. Azurite works cross-platform on Windows, Linux, and OS X. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,6 +5,21 @@
...
+# Important Update on Azurite Direction & Future Work
+
+Dear Azurite community, many thanks for your interest and support.  
+We thought we might be able to reduce some of our testing overhead by incorporating the Azure Storage Node tests in the dev pipeline, but that just showed us how far we really had to go.
+
+We were struggling to maintain parity with our APIs, community contributions were growing, as was our number of issues.  
+
+**To fix this**, we have been focusing on the completion of a new more easily maintainable architecture for Azurite, using Typescript and automatically generated interfaces.
+
+The plan is to implement blob storage, followed by table storage, then queues.  Each with persistent storage.
```
XiaoningLiu: Looks all good. The service priorities needed to be discussed. As we currently don't have a swagger for table storage.

seguler: Correct - we don't have a Table API in swagger; last time I checked it was impossible to describe them in swagger.

edwin-huber: OK, we can figure out an alternative to this if the other services are unable to produce swagger definitions, we can either :  
1. produce our own swagger definition based on the REST documentation, as our source of truth for the API.
2. create those API definitions manually ourselves, without the auto generation, in the same format as the blob API.

Does that sound workable to you?
Even if we work on the other APIs manually, we would still have the advantage of the blob API being auto-generated.

PR #175, opened 2019-03-22
README now includes shared key credentials

willKim19:
It was very difficult for me to find this information, I had to dig through a lot of other Azure documentation to find it.


edwin-huber:
Many thanks for catching and address this... as we work with Azure Storage Emulation all the time, this is not so obvious 👍

PR #176, opened 2019-03-28
NewArch : set Blob MetaData

edwin-huber:
Hi,

I refactored the blob api code a little, and implemented the get meta.
The other APIs will follow. 😸


src/blob/handlers/BlobHandler.ts
```diff
@@ -91,28 +79,16 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
     options: Models.BlobDeleteMethodOptionalParams,
     context: Context
   ): Promise<Models.BlobDeleteResponse> {
+    // Will throw exception first if blob does not exist before delete
+    // need to use ts-ignore to avoid compilation error.
```
XiaoningLiu: Why there will be a TS compile error?

src/blob/handlers/BlobHandler.ts
```diff
@@ -91,28 +79,16 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
+    // @ts-ignore
+    const blob = await this.getSimpleBlobFromStorage(context);
+
     const blobCtx = new BlobStorageContext(context);
     const accountName = blobCtx.account!;
```
XiaoningLiu: Because we already got `blob` instance by `getSimpleBlobFromStorage`. We can get `acccountName` and other names from `blob` instance directly.

```javascript
const accountName = blob.accountName;
const containerName = blob.containerName;
const blobName = blob.name;
```

src/blob/handlers/BlobHandler.ts
```diff
@@ -7,12 +7,10 @@ import Context from "../generated/Context";
...
 import { API_VERSION } from "../utils/constants";
 import BaseHandler from "./BaseHandler";
+import * as IBlobDataStore from "../persistence/IBlobDataStore";
 
 export default class BlobHandler extends BaseHandler implements IBlobHandler {
-  public async download(
-    options: Models.BlobDownloadOptionalParams,
-    context: Context
-  ): Promise<Models.BlobDownloadResponse> {
+  private async getSimpleBlobFromStorage(context: Context): Promise<any> {
```
XiaoningLiu: Wow, that's great! Help reducing lots of duplicated codes.

XiaoningLiu: Return type should be `Promise<BlobModel>`

src/blob/handlers/BlobHandler.ts
```diff
@@ -7,12 +7,10 @@ import Context from "../generated/Context";
 import IBlobHandler from "../generated/handlers/IBlobHandler";
 import { API_VERSION } from "../utils/constants";
 import BaseHandler from "./BaseHandler";
+import * as IBlobDataStore from "../persistence/IBlobDataStore";
```
XiaoningLiu: We prefer using `import {XXX, YYY } from Path` to load what we need, instead of using `import * as XXX`, which will import everything leading to potential lower performance.

When we do need load most of the exported objects from a file, we can use `import * as xxx`.

I think in this case, we can load what we need with `import {AAA, BBB}, CCC from Path`.

src/blob/handlers/BlobHandler.ts
```diff
@@ -143,7 +119,21 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
+    const blob = await this.getSimpleBlobFromStorage(context);
+    const blobCtx = new BlobStorageContext(context);
+
+    blob.metadata = options.metadata;
+
+    await this.dataStore.updateBlob(blob);
+
+    const response: Models.BlobSetMetadataResponse = {
+      statusCode: 200,
+      requestId: blobCtx.contextID,
```
XiaoningLiu: you can use `context.contextID`, so `blobCtx` can be removed.

src/blob/handlers/BlobHandler.ts
```diff
@@ -227,27 +217,16 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
-      accountName,
-      containerName,
-      blobName
-    );
-    if (!blob) {
-      throw StorageErrorFactory.getBlobNotFound(blobCtx.contextID!);
+    const blob = await this.getSimpleBlobFromStorage(context);
+    let blockBlob;
+    if (blob.properties.blobType === Models.BlobType.BlockBlob) {
+      blockBlob = <IBlobDataStore.BlobModel>blob;
```
XiaoningLiu: After updating return type of `getSimpleBlobFromStorage`, you will not need type cast here.

src/blob/handlers/BlobHandler.ts
```diff
@@ -227,27 +217,16 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
-    const blob = await this.dataStore.getBlob(
-      accountName,
-      containerName,
-      blobName
-    );
-    if (!blob) {
-      throw StorageErrorFactory.getBlobNotFound(blobCtx.contextID!);
+    const blob = await this.getSimpleBlobFromStorage(context);
+    let blockBlob;
+    if (blob.properties.blobType === Models.BlobType.BlockBlob) {
```
XiaoningLiu: Will this condition happen? downloadBlockBlob() will only be triggered from `download()` method, which already checks blobType.

src/blob/handlers/BlobHandler.ts
```diff
@@ -288,14 +267,19 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
-      const blocks = blob.committedBlocksInOrder;
-      bodyGetter = async () => {
-        return this.dataStore.readPayloads(
-          blocks.map(block => block.persistencyID),
-          rangeStart,
-          rangeEnd + 1 - rangeStart
-        );
-      };
+      const blocks = blockBlob.committedBlocksInOrder;
+      if (blocks != undefined) {
```
XiaoningLiu: This will happen? line 259 already checks `blob.committedBlocksInOrder === undefined `

tests/blob/apis/blob.test.ts
```diff
@@ -85,4 +85,14 @@ describe("BlobAPIs", () => {
   it("delete", async () => {
     await blobURL.delete(Aborter.none);
   });
+
+  it("setMetadata with new metadata set", async () => {
```
XiaoningLiu: Looks good!

XiaoningLiu:
@edwin-huber Thanks! Added some comments, please have a look.

src/blob/handlers/BlobHandler.ts
edwin-huber: Thanks. I reduced the scope of the import accordingly.

src/blob/handlers/BlobHandler.ts
edwin-huber: "noUnusedLocals": true in tsconfig

src/blob/handlers/BlobHandler.ts
edwin-huber: theoretically we could remove all the calls to the creation of a new storage context object, as I don't think there is a chance that something can change the object that we are using... right?

src/blob/handlers/BlobHandler.ts
edwin-huber: On my system I get a compile error if I do not check for an undefined object:

src/blob/handlers/BlobHandler.ts(269,11): error TS2532: Object is possibly 'undefined'.

edwin-huber: To expand on this, if I do not check for undefined, I get a compile error:
```shell
TSError: ⨯ Unable to compile TypeScript:
src/blob/handlers/BlobHandler.ts(269,11): error TS2532: Object is possibly 'undefined'.
```
so I created the check and appropriate responses
```typescript
const blocks = blockBlob.committedBlocksInOrder;
      if (blocks != undefined) {
```
committedBlocksInOrder can return undefined, so we either need to ignore the compile error here, or check for undefined.  
What do you think?

src/blob/handlers/BlobHandler.ts
edwin-huber: if the const is not checked at assignment, it causes a compile error due to strict type / null checking.

TS2532: Object is possibly 'undefined'

src/blob/handlers/BlobHandler.ts
edwin-huber: now refactored

PR #128, opened 2018-12-03
Add support for HTTP Method Overrides

gpatryk:
Fixes #127.

PR #184, opened 2019-04-05
Support empty RowKey

### LINKED ISSUE #183: Empty RowKey is not supported in Retrieve table operation.
Formally, RowKey with empty string as value is legal.
Actually when RowKey is passed as empty string the Entity Retrieve operation is handled as Query Entities operation and despite the PartitionKey is passed the response is an array.

AleF83:
This PR fixes the bug when in Retrieve Entity Operation the parameters are PartitionKey with some value and RowKey with empty string value.

Fixes #183


AleF83:
The PR contains #128. So it should be merged after it

edwin-huber:
We have confirmed the behaviors in the service and will support X-HTTP-method 😄   
I'm now merging this into legacy-dev  
We should release a new update to the legacy branch soon

Issue #179, opened 2019-03-29
[NewArch][API][Blob] Get Blob Metadata


edwin-huber:
Needs update of the autogenerated interface, as currently there is no blob.getMetadata

Issue #161, opened 2019-03-11
[NewArch][API][Blob] List Blobs


vinjiang:

XiaoningLiu:
@pdressel There are 2 list blobs handler APIs to implement: `listBlobHierarchySegment` and `listBlobFlatSegment`. I implemented `listBlobHierarchySegment` for testing purpose. Just notice you.

PR #188, opened 2019-04-11
Newarch dev edwin phase 2

edwin-huber:
Addresses:
1. #161 
1. #178 
1. #179 

Implemented List Flat Blobs  
Added tests for setting headers  
need to extend implementation and tests to cover the use of headers with any set / get request


src/blob/handlers/ContainerHandler.ts
```diff
@@ -205,7 +205,67 @@ export default class ContainerHandler extends BaseHandler
...
+      options.maxresults,
+      marker
+    );
+
+    const blobItems: Models.BlobItem[] = [];
+    const blobPrefixes: Models.BlobPrefix[] = [];
+    const blobPrefixesSet = new Set<string>();
+    const prefixLength = options.prefix.length;
+
+    for (const blob of blobs) {
```
XiaoningLiu: `this.dataStore.listBlobs()` has filtered `prefix`, don't need to filter again here

src/blob/handlers/ContainerHandler.ts
```diff
@@ -205,7 +205,67 @@ export default class ContainerHandler extends BaseHandler
...
+      }
+      const prefixPos = blob.name.indexOf(options.prefix);
+      // This is a blob with matching prefix
+      if (prefixPos >= 0) {
+        blob.deleted = blob.deleted !== true ? undefined : true;
+        blobItems.push(blob);
+      }
+    }
+
+    const iter = blobPrefixesSet.values();
```
XiaoningLiu: blobPrefixesSet doesn't any any values.

src/blob/handlers/ContainerHandler.ts
```diff
@@ -205,7 +205,67 @@ export default class ContainerHandler extends BaseHandler
...
+      date: context.startTime,
+      serviceEndpoint,
+      containerName,
+      prefix: options.prefix,
+      marker: options.marker,
+      maxResults: options.maxresults || this.LIST_BLOBS_MAX_RESULTS_DEFAULT,
+      delimiter,
+      segment: {
+        blobItems,
+        blobPrefixes
```
XiaoningLiu: For `ContainerListBlobFlatSegmentResponse`, `blobPrefixes` is not needed. Can we put empty array to `blobPrefixes`?

src/blob/handlers/ContainerHandler.ts
```diff
@@ -205,7 +205,67 @@ export default class ContainerHandler extends BaseHandler
     options: Models.ContainerListBlobFlatSegmentOptionalParams,
     context: Context
   ): Promise<Models.ContainerListBlobFlatSegmentResponse> {
-    throw new NotImplementedError(context.contextID);
+    const container = await this.getSimpleContainerFromStorage(context);
+    const request = context.request!;
```
XiaoningLiu: `request` variable seems useless

XiaoningLiu:
@edwin-huber Thanks Edwin! That's really great! 

I Added some comments, please have a look!

src/blob/handlers/BlobHandler.ts
```diff
@@ -125,11 +125,48 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
     throw new NotImplementedError(context.contextID);
   }
 
+  // see also https://docs.microsoft.com/en-us/rest/api/storageservices/set-blob-properties
```
XiaoningLiu: Can be improved to use standard JSDoc comments format.

You may want to install Visual Studio Code Plugin "Document This", which could help generate standard JSDoc comments for methods, classes, interfaces. After installing, goto previous line of code you want to comment, type "/**" then press "Enter", all comments generated!

XiaoningLiu: Other Visual Studio Code plugins I'm using to help automatically code fix and format:
* tslint
* prettier
* TypeScript Import Sorter

edwin-huber: Done

src/blob/handlers/ContainerHandler.ts
edwin-huber: removed prefixes processing from flat list logic

src/blob/handlers/ContainerHandler.ts
edwin-huber: removed prefixes

PR #203, opened 2019-04-17
Snapshot implementation for #190

edwin-huber:
Snapshot API has been implemented, and the DB schema updated.
#190 
There are now issues with the correctness of several APIs and tests which need to be corrected to account for snapshots in the blob collection.

1. collection indices extended by snapshot property.
1. listBlobs API on persitent storage extended to account for snapshots

### Open tasks:

1. All special naming tests now failing
1. Correctness of list blob tests needs to be checked.
Number of expected results no longer correct on flat and hierarchy tests.
1. Correctness of delete / undelete operations 
BlobDeleteMethodOptionalParams.deleteSnapshots
Required if the blob has associated snapshots. Specify one of the following two options:
 include: Delete the base blob and all of its snapshots. 
only: Delete only the blob's snapshots and not the blob itself. 
Possible values include: 'include', 'only'
Means that we need to check for snapshots and then process the options accordingly.
Need to check how the service API deals with missing required param at this stage.

### DIFF
--- a/src/blob/handlers/BlobHandler.ts
+++ b/src/blob/handlers/BlobHandler.ts
@@ -757,6 +757,7 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
 
   /**
    * create snapshot
+   * https://docs.microsoft.com/en-us/rest/api/storageservices/snapshot-blob
    *
    * @param {Models.BlobCreateSnapshotOptionalParams} options
    * @param {Context} context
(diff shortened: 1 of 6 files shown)


XiaoningLiu:
Seems good for snapshot implementation. Failed test cases caused by list blob is not fully updated.

XiaoningLiu:
@edwin-huber  We may need to pass existing test cases before merging the PR

PR #214, opened 2019-05-09
Add Dockerfile, remove package-lock to avoid issues with stale modules

tomhosking:
Added example Dockerfile. Also removed `package-lock.json` as this should be generated when `npm install` is run.


Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# Blob Storage Emulator
+EXPOSE 10000
+# Azure Queue Storage Emulator
+EXPOSE 10001
+# Azure Table Storage Emulator
+EXPOSE 10002
+
+
+CMD ["azurite","-l", "/azurite-data", "--blobHost", "0.0.0.0"]
```
XiaoningLiu: Add a space between `"azurite",` and `"-l"`

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM node:10.6-alpine
+
+WORKDIR /opt/azurite
+
+COPY package.json package.json
```
XiaoningLiu: Should include `package.json.lock` before `npm install`

XiaoningLiu:
Please resolve the conflict and comments before we can merge the PR.

Dockerfile
tomhosking: Added command to copy `package-lock.json` to container (and readded it to repo)

Dockerfile
tomhosking: Done

PR #247, opened 2019-08-05
Fix listblobs order when filtering by prefix

jeffsaracco:
In the Azure docs for List Blobs it specifies

> Blobs are listed in alphabetical order in the response body, with upper-case letters listed first.

(via: https://docs.microsoft.com/en-us/rest/api/storageservices/list-blobs search for "alphabetical")

However we have found that Azurite does not guarantee that ordering when listing blobs with a `prefix` option. This PR simply adds a test that exposes that behavior and adds a lokijs `simplesort` call to the `listBlobs` function.

We found this bug by trying to use Azurite for testing when implementing a client app for Azure Blob storage. When we went to write paging using the `Marker` functionality we noticed that when the prefix option is sent to Azurite the results are returned in reverse create order, not alphabetical order. When tested against `<account>.blob.core.microsoft.net` the order is correct in all scenarios.

### DIFF
--- a/src/blob/persistence/LokiBlobDataStore.ts
+++ b/src/blob/persistence/LokiBlobDataStore.ts
@@ -447,7 +447,7 @@ export default class LokiBlobDataStore implements IBlobDataStore {
     blob?: string,
     prefix: string | undefined = "",
     maxResults: number | undefined = 5000,
-    marker?: string,
+    marker: string = "",
     includeSnapshots?: boolean | undefined
   ): Promise<[T[], string | undefined]> {
     const query: any = {};
@@ -462,10 +462,6 @@ export default class LokiBlobDataStore implements IBlobDataStore {
     }
     if (container !== undefined) {
       query.containerName = container;
-    }
-
-    if (marker === undefined) {
-      marker = "";
     }
 
     const coll = this.db.getCollection(this.BLOBS_COLLECTION);
--- a/src/blob/persistence/LokiBlobDataStore.ts
+++ b/src/blob/persistence/LokiBlobDataStore.ts
@@ -502,7 +502,7 @@ export default class LokiBlobDataStore implements IBlobDataStore {
     if (docs.length < maxResults) {
       return [docs, undefined];
     } else {
-      const nextMarker = docs.length;
+      const nextMarker = docs.length + marker;
       return [docs, nextMarker];
     }
   }
--- a/src/blob/persistence/LokiBlobDataStore.ts
+++ b/src/blob/persistence/LokiBlobDataStore.ts
@@ -464,7 +464,9 @@ export default class LokiBlobDataStore implements IBlobDataStore {
       query.containerName = container;
     }
 
-    query.$loki = { $gt: marker };
+    if (marker === undefined) {
+      marker = 0;
+    }
 
     const coll = this.db.getCollection(this.BLOBS_COLLECTION);
 
@@ -474,6 +476,7 @@ export default class LokiBlobDataStore implements IBlobDataStore {
         .chain()
         .find(query)
         .simplesort("name")
+        .offset(marker)
         .limit(maxResults)
         .data();
     } else {
@@ -484,6 +487,7 @@ export default class LokiBlobDataStore implements IBlobDataStore {
           return obj.snapshot.length === 0;
         })
         .simplesort("name")
+        .offset(marker)
         .limit(maxResults)
         .data();
     }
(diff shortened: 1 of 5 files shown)


jeffsaracco:
I should also add that I was able to recreate this in LokiJS itself, in the [sandbox](https://rawgit.com/techfort/LokiJS/master/examples/sandbox/LokiSandbox.htm), but I couldn't find any documentation specifying item order there. Since the Azure docs _do_  specify an order to return, I figured this is the best place to make the fix.

The script I used for recreating the in the Loki sandbox is below.

Loki Sandbox script:
```javascript
var db = new loki('sandbox.db');

// Add a collection to the database
var items = db.addCollection('items');

// Add some documents to the collection
items.insert({ name : 'mjolnir', owner: 'thor', maker: 'dwarves' });
items.insert({ name : 'draupnir', owner: 'odin', maker: 'elves' });
items.insert({ name : 'gungnir', owner: 'odin', maker: 'elves' });
items.insert({ name : 'tyrfing', owner: 'Svafrlami', maker: 'dwarves' });

// These statements send to Text Output
logText('odins items');
logObject(items.find({ 'owner': 'odin' }));

function makerView(obj){
  return obj.maker === 'elves';
}

logText('regex');
var createOrder = items.chain().find({'owner': {$regex: '^odi*'}}).data();
var reverseLexOrder = items.chain().find({'owner': {$regex: '^odi*'}}).where(makerView).data();
var fixLexed = items.chain().find({'owner': {$regex: '^odi*'}}).where(makerView).simplesort('name').data();
logObject(createOrder);
logObject(reverseLexOrder);
logObject(fixLexed);

// This statement sends to Inspector
inspectObject(db);
```


If this issue is something that should be fixed within Loki itself I'm happy to work on that instead.

XiaoningLiu:
@jeffsaracco Thanks! We need a new `nextMarker` generation logic if we sort by names instead of by `$loki` id (doc insert order). Currently, the `nextMarker` is selected from the last element in a segement, and the last element always has the max $loki in this segment. If we sort by name, then this guarantee will not exist. Any good thoughts?

```TypeScript
    if (docs.length < maxResults) {
      return [docs, undefined];
    } else {
      const nextMarker = docs[docs.length - 1].$loki;
      return [docs, nextMarker];
    }
```

WillAbides:
@XiaoningLiu I've been working with @jeffsaracco on this.  `nextMarker` is broken in master.  That's what originally led us to find this issue. On master, when you filter by prefix it reverses order and the nextMarker is the first item it returns. So when you make a second request with that marker it returns the same results as the first request minus the first result.  I know that explanation isn't completely clear.  Jeff is working on adding a test that demonstrates this.

Using this branch ended up solving the nextMarker issue for us.

jeffsaracco:
@XiaoningLiu Good catch, I hadn't thought of the `nextMarker` in my original patch. It did break it. However I think I have a fix now using `offset` with `limit`. The next marker no longer uses `$loki` but the last offset returned. Since the collection has an index on `name` this shouldn't impact speed too much (if at all)

XiaoningLiu:
Hi @jeffsaracco @WillAbides Really thanks, it's alomost there! Using offset will work in most of the times. But the offset number could "shift" when creating a blob with a "smaller" name. For example.

We have 4 blobs:

b002, b001, b004, b003

When first list blob request comes in. We will order it and select first 2 blobs b001 & b002 as response. And the offset (nextMarker) will be 2.

b001, b002, b003, b004

Before next list blob request comes in, we may create 2 new blobs a001, a002. So we have 6 blobs here:

b002, b001, b004, b003, a001, a002

Then next listBlobs request comes in with nextMarker 2. We sort the blobs, and select from offset 2. However b001 and b002 will be returned, which already been returned in first list blob response.

a001, a002, b001, b002, b003, b004

I'm thinking whether we can use blobName as next marker when we order blobs. We need to change nextMarker from number to string type in `listBlobs` resposne values. Any ideas?

jeffsaracco:
@XiaoningLiu I had thought of that, but didn't want to change the function signature - if that's ok with you, I'm happy to make that change as well.

I'll update this PR with that change

src/blob/persistence/LokiBlobDataStore.ts
```diff
@@ -447,9 +447,9 @@ export default class LokiBlobDataStore implements IBlobDataStore {
     blob?: string,
     prefix: string | undefined = "",
     maxResults: number | undefined = 5000,
-    marker?: number | undefined,
+    marker?: string,
```
XiaoningLiu: We can provide `marker`  parameter a default value `""`, then don't need following assignment

```javascript
    if (marker === undefined) {
      marker = "";
    }
```

jeffsaracco:
@XiaoningLiu thank you for the approval - I made your suggestion as well.

Issue #275, opened 2019-11-08
list_containers does not sort the results


timuralp:
### Which service(blob, file, queue, table) does this issue concern?
Blob

### Which version of the Azurite was used?
3.2

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
Dockerhub

### What's the Node.js version?
10.15.3

### What problem was encountered?
list_blobs returns the response not sorted in order:
Here is an example output I get using the Python SDK:
```
>>> [cont.name for cont in c.list_containers()]
[u'test-blob-\xc3\xaf', 'test', 'test2', 'a', 'z', 'e']
```

The behavior is documented [here](https://docs.microsoft.com/en-us/rest/api/storageservices/list-containers2): Containers are listed in alphabetical order in the response body.

### Steps to reproduce the issue?
Create containers not in alphabetical order, e.g. container "foo" then container "bar", then list containers.

### Have you found a mitigation/solution?
No

XiaoningLiu:
Hi @timuralp Thanks! Will update for a fix.

sangeetha5491:
Can i work on this?

XiaoningLiu:
@sangeetha5491 Yes! As list blobs already have order, you can do this for containers to. Notice that, we have 2 metadata store implemntation, one based on loki and another for SQL. Please update them at the same time.

sangeetha5491:
thanks! Im working on it

sangeetha5491:
@XiaoningLiu https://github.com/Azure/Azurite/pull/280

XiaoningLiu:
@sangeetha5491 Thanks! Few comments added.

edwin-huber:
Fix was merged, closing issue.

PR #280, opened 2019-11-15
List containers should be displayed in sorted order

sangeetha5491:

Links to issue https://github.com/Azure/Azurite/issues/275

### PR Branch Destination

- For Azurite V3, please send PR to `dev` branch.

### Always Add Test Cases

### Development Guideline

### DIFF
--- a/src/blob/persistence/LokiBlobMetadataStore.ts
+++ b/src/blob/persistence/LokiBlobMetadataStore.ts
@@ -319,6 +319,7 @@ export default class LokiBlobMetadataStore
       .chain()
       .find(query)
       .limit(maxResults + 1)
+      .simplesort("name")
       .data();
 
     if (docs.length <= maxResults) {
--- a/src/blob/persistence/LokiBlobMetadataStore.ts
+++ b/src/blob/persistence/LokiBlobMetadataStore.ts
@@ -293,8 +293,8 @@ export default class LokiBlobMetadataStore
    * @param {string} account
    * @param {string} [prefix=""]
    * @param {number} [maxResults=5000]
-   * @param {number} [marker=0]
-   * @returns {(Promise<[ContainerModel[], number | undefined]>)}
+   * @param {string} [marker=""]
+   * @returns {(Promise<[ContainerModel[], string | undefined]>)}
    * @memberof LokiBlobMetadataStore
    */
   public async listContainers(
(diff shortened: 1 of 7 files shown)


sangeetha5491:
build

sangeetha5491:
How do I find out why the 4 tests were canceled?

sangeetha5491:
@XiaoningLiu I can't see your comments.

blueww:
@sangeetha5491
1. Please add the change log for the change in: https://github.com/Azure/Azurite/blob/master/ChangeLog.md, you can add it under "Upcoming Release" tag
2. Consider the change for marker  in adding test case.

sangeetha5491:
@XiaoningLiu and @blueww I have addressed the review comments. Please review it again.

src/blob/persistence/LokiBlobMetadataStore.ts
```diff
@@ -293,32 +293,32 @@ export default class LokiBlobMetadataStore
...
+    marker: string = ""
+  ): Promise<[ContainerModel[], string | undefined]> {
     const coll = this.db.getCollection(this.CONTAINERS_COLLECTION);
 
     const query =
       prefix === ""
-        ? { $loki: { $gt: marker }, accountName: account }
+        ? { name: { $gt: marker }, accountName: account }
         : {
             name: { $regex: `^${this.escapeRegex(prefix)}` },
```
XiaoningLiu:
```suggestion
            name: { $regex: `^${this.escapeRegex(prefix)}`, $gt: marker },
```

tests/blob/apis/service.test.ts
```diff
@@ -142,6 +142,104 @@ describe("ServiceAPIs", () => {
...
+    await containerURL3.create(Aborter.none);
+    const result = await serviceURL.listContainersSegment(
+      Aborter.none,
+      containerName2,
+      { maxresults: 1 }
+    );
+    assert.equal(result.containerItems.length, 1);
+    assert.ok(result.containerItems[0].name.startsWith("bb"));
+    assert.equal(result.nextMarker, containerName3);
+    await containerURL1.delete(Aborter.none);
```
XiaoningLiu: Can we use containerName3 as the marker do the query again? Then we will get containerName1 in the response with nextMarker undefined.

Others in the PR looks good.

sangeetha5491: nextMarker contains value of the last container in the previous list. When we fetch the list of containers, it gets fetches the list after the nextMarker. So the response will have nextMarker as containerName1.

sangeetha5491: I'm adding a test case, it looks like it does not set nextMarker to undefined for Loki.

sangeetha5491: done

Issue #293, opened 2019-12-13
Get/download blob snapshot fail with 404


blueww:
### Which service(blob, file, queue, table) does this issue concern?
blob
### Which version of the Azurite was used?
3.3.0-preview
### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
VS code
### What's the Node.js version?
v10.15.3
### What problem was encountered?
Get/download blob snapshot always fail with 404, while Create snapshot works well, list also include snapshot. 
Test with Powershell, which based on .NET Storage client library v11.0
### Steps to reproduce the issue?
create snapshot of a blob, them try to download it.

> 2019-12-12T04:02:39.107Z 399cb32e-d412-4f2d-b402-674adf0cf429 info: HandlerMiddleware: DeserializedParameters={"options":{"snapshot":"2019-12-12T04:02:19.0110000Z","leaseAccessConditions":{},"cpkInfo":{},"modifiedAccessConditions":{}}}
> 2019-12-12T04:02:39.108Z 399cb32e-d412-4f2d-b402-674adf0cf429 error: ErrorMiddleware: Received a MiddlewareError, fill error information to HTTP response
[...]

PR #297, opened 2019-12-13
Fix get snapshot 404 issue, and add API version validation

blueww:
1. Fix issue: https://github.com/Azure/Azurite/issues/293
2. Add API version validation , validate API version get from https://docs.microsoft.com/en-us/rest/api/storageservices/previous-azure-storage-service-versions

Don't have test now since:
1. For snapshot, JS test can't repro the issue. The issue is from PSH with XSCL
2. For API version, it's not easy to add test. We have manually test with PSH. Opened a task (P2) to add it later.

### PR Branch Destination

- For Azurite V3, please send PR to `dev` branch.

### Always Add Test Cases

### Development Guideline

### DIFF
--- a/src/blob/context/blobStorageContext.middleware.ts
+++ b/src/blob/context/blobStorageContext.middleware.ts
@@ -2,11 +2,13 @@ import { NextFunction, Request, Response } from "express";
 import uuid from "uuid/v4";
 
 import logger from "../../common/Logger";
+import { checkApiVersion } from "../../common/utils/utils";
 import StorageErrorFactory from "../errors/StorageErrorFactory";
 import {
   DEFAULT_CONTEXT_PATH,
   HeaderConstants,
   SECONDARY_SUFFIX,
+  ValidAPIVersions,
   VERSION
 } from "../utils/constants";
 import BlobStorageContext from "./BlobStorageContext";
@@ -26,11 +28,16 @@ export default function blobStorageContextMiddleware(
 ): void {
   // Set server header in every Azurite response
   res.setHeader(HeaderConstants.SERVER, `Azurite-Blob/${VERSION}`);
+  const requestID = uuid();
+
+  const apiVersion = req.header(HeaderConstants.X_MS_VERSION);
+  if (apiVersion !== undefined) {
+    checkApiVersion(apiVersion, ValidAPIVersions, requestID);
+  }
 
   const blobContext = new BlobStorageContext(res.locals, DEFAULT_CONTEXT_PATH);
   blobContext.startTime = new Date();
 
-  const requestID = uuid();
   blobContext.xMsRequestID = requestID;
 
   logger.info(
@@ -76,9 +83,7 @@ export default function blobStorageContextMiddleware(
     );
 
     logger.error(
-      `BlobStorageContextMiddleware: BlobStorageContextMiddleware: ${
-        handlerError.message
-      }`,
+      `BlobStorageContextMiddleware: BlobStorageContextMiddleware: ${handlerError.message}`,
       requestID
     );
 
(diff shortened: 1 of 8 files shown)


src/blob/context/blobStorageContext.middleware.ts
```diff
@@ -26,11 +28,14 @@ export default function blobStorageContextMiddleware(
 ): void {
   // Set server header in every Azurite response
   res.setHeader(HeaderConstants.SERVER, `Azurite-Blob/${VERSION}`);
+  const requestID = uuid();
+
+  const apiVersion = String(req.header("x-ms-version"));
```
XiaoningLiu: Should move "x-ms-version" into "constants.ts" `HeaderConstants`;

Should check the situation when `x-ms-version` is undefined, when we will fallback to Azurite default API Version.

src/common/utils/utils.ts
```diff
@@ -1,8 +1,33 @@
...
   return date1 > date2 ? date2 : date1;
 }
+
+export function checkApiVersion(
+  inputApiVersion: string,
+  validApiVersions: Array<string>,
+  requestId: string
+): void {
+  if (
+    inputApiVersion !== undefined &&
```
XiaoningLiu: parameter `inputApiVersion` is required in this definition. If you want to allow it tobe optional, you can update the type from `string` to `string | undefined`.

src/blob/context/blobStorageContext.middleware.ts
XiaoningLiu:
```suggestion
  const apiVersion = req.header("x-ms-version");
  if (typeof apiVersion === "string") {
    checkApiVersion(apiVersion, ValidAPIVersion, requestID);
  }
```

src/blob/utils/constants.ts
```diff
@@ -59,3 +59,24 @@ export const DEFAULT_BLOB_PERSISTENCE_ARRAY: StoreDestinationArray = [
     maxConcurrency: 10
   }
 ];
+
+export const ValidAPIVersion = [
```
XiaoningLiu:
```suggestion
export const ValidAPIVersions = [
```

src/queue/context/queueStorageContext.middleware.ts
```diff
@@ -28,13 +30,17 @@ export default function queueStorageContextMiddleware(
   // Set server header in every Azurite response
   res.setHeader(HeaderConstants.SERVER, `Azurite-Queue/${VERSION}`);
 
+  const requestID = uuid();
+
+  const apiVersion = String(req.header("x-ms-version"));
```
XiaoningLiu: same comments with blob.

src/queue/utils/constants.ts
```diff
@@ -88,3 +88,24 @@ export const DEFAULT_QUEUE_PERSISTENCE_ARRAY: StoreDestinationArray = [
     maxConcurrency: 1
   }
 ];
+
+export const ValidAPIVersion = [
```
XiaoningLiu:
```suggestion
export const ValidAPIVersions = [
```

src/blob/utils/constants.ts
blueww: Fixed

src/queue/context/queueStorageContext.middleware.ts
blueww: Has put "x-ms-version" to constants

src/blob/context/blobStorageContext.middleware.ts
blueww: Has put "x-ms-version" to constants

Has check undefined `x-ms-version`,

src/common/utils/utils.ts
blueww: Has check undefined x-ms-version outside. and remove check for undefined inside.

src/queue/utils/constants.ts
blueww: fixed

Issue #378, opened 2020-01-03
SharedKeyLite support for blob service


shivangag:
### Which service(blob, file, queue, table) does this issue concern?
Blob service

### Which version of the Azurite was used?
Azurite v3

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
Github master branch

blueww:
@shivangag 

Thanks for the request!

Would you please share the reason to use SharedKeyLite?
Is there any reason blocks you from switching to SharedKey?

shivangag:
@blueww, I am using [JCloud](https://jclouds.apache.org/) to create a multi cloud app. They currently only support SharedKeyLite based authorization for azure.
Therefore, for local testing it would be great if Azurite support it.

blueww:
@shivangag 
I might need more time to investigate/discuss this issue, and will update this issue when get any result.

shivangag:
@blueww, Is there any design reason to not include SharedKeyLite in blob service?
As I can see [here](https://github.com/Azure/Azurite/blob/6d444ef1ab6f2933d63ca5fffa07de3abffa8f12/src/queue/authentication/QueueSharedKeyAuthenticator.ts#L52) the queue service already supports it. I can open a PR with required changes following the queue service implementation.
Will that be fine?

XiaoningLiu:
Hi @shivangag PR is welcome!

PR #402, opened 2020-02-12
Add SharedKeyLite auth support to blob service

shivangag:
PR for issue #378.

### DIFF
--- a/src/blob/authentication/BlobSharedKeyAuthenticator.ts
+++ b/src/blob/authentication/BlobSharedKeyAuthenticator.ts
@@ -49,24 +49,16 @@ export default class BlobSharedKeyAuthenticator implements IAuthenticator {
       );
     }
 
+    const authType = authHeaderValue.split(" ")[0] as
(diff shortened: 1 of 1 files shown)


src/blob/authentication/BlobSharedKeyAuthenticator.ts
```diff
@@ -49,24 +49,16 @@ export default class BlobSharedKeyAuthenticator implements IAuthenticator {
       );
     }
 
+    const authType = authHeaderValue.split(" ")[0] as
```
XiaoningLiu: What if an invalid word besides "SharedKey" or "SharedKeyLite"? Better to stop auth for any invalid keyword.

blueww:
@shivangag 
Would you please add test cases to validate SharedKeylite works on both blob and Queue?
Besides that, Please sign the cla license.

vinjiang:
We would love to see community contribution. Please follow the contribution guideline to engage with us.

Issue #403, opened 2020-02-12
Optionally support production style URL format http://accountname.localhost/container


ayeltsov:
We have extensive and multi-layered code base which assumes production style URL format 
https://accountname.core.windows.net/container

We have looked what would it take to support emulated format 
http://localhost/accountname/container and due to code layout and assumptions of which part of URL means what in various layers, it would be extensive and error-prone.

This ask is to optionally support production style URL format
http://accountname.localhost/container

We will take care of name resolution, say via hosts file, or via local DNS server so that accountname.localhost resolves to 127.0.0.1

But we need help in routing layers so that URL parts are properly recognized if some flag is set (e.g. set AZURITE_URL_STYLE=PRODUCTION, or set AZURITE_USE_PRODUCTION_URL_STYLE=1)

Can you please send some code pointers of which places we'd need to change so that we can prototype this change? 
Thanks!

XiaoningLiu:
Hi @ayeltsov Personally like the idea to have a flag for switch!

Please refer to this where Azurite parser account, container, and blob names from request.

https://github.com/Azure/Azurite/blob/be4dc2c9d44e3ba643e616795c26cb48dedd5c8e/src/blob/middlewares/blobStorageContext.middleware.ts#L54

queue service has a similar queueStorageContext.middleware.ts too.

XiaoningLiu:
There are some further updates after PR https://github.com/Azure/Azurite/pull/416 @ayeltsov likes to have following updates.

tests
documentation
Some configuration options:
add custom domain name (for example, if someone wants to mock .blob.core.windows.net, they should be able to do this)
also considering to add option to any domain suffix (e.g. http[s]://account.*/container

PR #416, opened 2020-03-03
Support production-style URLs (e.g. http://account.localhost/container)

ayeltsov:
Solves #403 - adds ability to use production-style URLs when using "account.localhost" as the host name.
E.g. default URL format for emulator is

> http://hostname[:port]/account/container

this adds ability to use 

> http[s]://account.localhost[:port]/container

It's users' responsibility that account.localhost resolves to 127.0.0.1. On Windows (I believe this works out of the box on Linux), this can be achieved by adding hosts file record, for example. 

C:\Windows\System32\drivers\etc\hosts
```
<...>
127.0.0.1       devstoreaccount1.localhost

```

I contemplated putting this functionality under some sort of flag (like environmental variable).
But then I realized that it's very unlikely to break any existing use patterns, as it requires issuing API requests to accountname.localhost, and having that resolve to 127.0.0.1, hence I proposed to enable this feature by default.

### DIFF
--- a/src/blob/handlers/BlobHandler.ts
+++ b/src/blob/handlers/BlobHandler.ts
@@ -576,7 +576,7 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
       sourceAccount,
       sourceContainer,
       sourceBlob
-    ] = extractStoragePartsFromPath(url.pathname);
+    ] = extractStoragePartsFromPath(url.hostname, url.pathname);
     const snapshot = url.searchParams.get("snapshot") || "";
 
     if (
--- a/src/blob/middlewares/blobStorageContext.middleware.ts
+++ b/src/blob/middlewares/blobStorageContext.middleware.ts
@@ -2,10 +2,10 @@ import { NextFunction, Request, Response } from "express";
 import uuid from "uuid/v4";
 
 import logger from "../../common/Logger";
+import { PRODUCTION_STYLE_URL_HOSTNAME } from "../../common/utils/constants";
 import { checkApiVersion } from "../../common/utils/utils";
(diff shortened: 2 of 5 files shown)


blueww:
@ayeltsov 
It seems only blob is supported.
Would you like to also add support for queue, and add related cases?

ayeltsov:
I'm going to work on this a bit more.
- tests
- documentation
- Some configuration options:
  - add custom domain name (for example, if someone wants to mock .blob.core.windows.net, they should be able to do this)
  - also considering to add option to any domain suffix (e.g. http[s]://account.*/container

XiaoningLiu:
> I'm going to work on this a bit more.
> 
> * tests
> * documentation
> * Some configuration options:
>   
>   * add custom domain name (for example, if someone wants to mock .blob.core.windows.net, they should be able to do this)
>   * also considering to add option to any domain suffix (e.g. http[s]://account.*/container

Hi @ayeltsov, I'd like to merge this PR for v3.6.0 planned in this week. Also looking forward to the further more updates mentioned! You can have another PR for that. Using GitHub issue #403 tracking this.

So . You can have the further updates

ayeltsov:
To be honest, I think I have initiated pull a bit prematurely. 
It worked for my limited scenario, but when I started experimenting with Storage Explorer, I noticed that for "localhost" endpoint suffix (e.g. AccountName=mystorage;AccountKey=TXlLZXk=;EndpointSuffix=localhost;DefaultEndpointsProtocol=http;), it sends requests to mystorage.blob.localhost:80

I did not account for that in my change.
I don't think we have to undo the change - it is benign - but I will definitely need to improve on it.

Thanks!

PR #437, opened 2020-04-24
Fix class casing

jongio:
I just noticed that casing was incorrect on one of the class names, so if someone copy and pastes the code it will not work.


blueww:
@XiaoningLiu 
The change looks good to me. 
Do you have any comments?
If not, we might can merge it.

PR #462, opened 2020-06-30
Adds test that listBlobs should only return committed blobs

dhadka:
This PR adds a test to reproduce issue https://github.com/Azure/Azurite/issues/463, where `listBlobs` does not support all possible includes options.  For example, if there are uncommitted blobs, then `listBlobs` is returning those uncommitted blobs with a length of 0.  This is a deviation from the actual Azure implementation, where `listBlobs` will filter out uncommitted blobs unless they are explicitly requested.


dhadka:
Closing, see the fix in https://github.com/Azure/Azurite/pull/464

Issue #463, opened 2020-06-30
listBlobs does not support all includes options, including uncommittedblobs


dhadka:
### Which service(blob, file, queue, table) does this issue concern?

Blob

### Which version of the Azurite was used?

3

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)

NuGet

### What's the Node.js version?

Any

### What problem was encountered?

The `listBlobs` method only respects the `includes: snapshot` option.  However, it does not respect `uncommittedblobs` or other options.  This is a discrepancy between Azurite and real Azure storage.

For example, `listBlobs` should only return committed blobs unless the `uncommittedblobs` options is specified.  Instead, Azurite lists all blobs, both committed and uncommitted.  This makes testing with Azurite difficult since unexpected blobs can appear in the container.

### Steps to reproduce the issue?

1. Stage a block without committing a blob.
2. List the blobs.

### Have you found a mitigation/solution?

Proposed fix in https://github.com/Azure/Azurite/pull/464

XiaoningLiu:
PR merged. Thanks @dhadka!

PR #464, opened 2020-06-30
listBlobs should only list committed blobs

dhadka:
Fixes https://github.com/Azure/Azurite/issues/463

Currently, `listBlobs` will return all blobs, including staged / uncommitted blobs.  This is a discrepancy from real Azure storage, where only committed blobs are returned unless you explicitly request them using the `includes: uncommittedblobs` option.


patrickcarnahan:
@XiaoningLiu quick turnaround! How do we know when this is available in docker hub?

XiaoningLiu:
We planned a release before next week.

Issue #461, opened 2020-06-28
[syncCopyFromURL] BlobNotFound - The specified blob does not exist.


AlexanderYW:
### Which service(blob, file, queue, table) does this issue concern?
blob

### Which version of the Azurite was used?
3.7.0

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
npm

### What's the Node.js version?
v12.13.1

### What problem was encountered?
[syncCopyFromURL](https://azuresdkdocs.blob.core.windows.net/$web/javascript/azure-storage-blob/12.1.2/classes/blobclient.html#synccopyfromurl) api is not working?, i'm trying to copy a blob but my test are failing when using Azurite, my test's works as expected when using AzureStorageEmulator

### Steps to reproduce the issue?

1. git clone https://github.com/AlexanderYW/Adonis-Drive-Azure-Storage.git
2. git checkout actions/blob-storage-emulator 
3. npm install
4. node test/blob.spec.js

### Error details
Details about this failing https://github.com/AlexanderYW/Adonis-Drive-Azure-Storage/pull/7/checks?check_run_id=816053411

### Have you found a mitigation/solution?
Not yet

XiaoningLiu:
Hi @AlexanderYW Can you enable Azurite debug log and share it here? SyncCopyBlob is not supported by Azurite yet. Like to check Azurite log for double confirm.

AlexanderYW:
Hi @XiaoningLiu, Sure I will provide you with the log in about 2 hours.

I'm not 100% sure that it's the SyncCopyBlob, just what I assumed it was since that's the main difference between the rest of the test and the ones that fails :)

AlexanderYW:
Hi @XiaoningLiu,

Here is the debug log
https://gist.github.com/AlexanderYW/7d20871b786f604a54eaa0f8d3ae9ffc

XiaoningLiu:
@AlexanderYW It's true it's an not implemented API. You can find "Current API is not implemented yet." in debug.log and response.

2020-06-29T14:57:53.995Z 4d778175-3e41-48d0-9b27-ee2e8d2dbeee info: BlobStorageContextMiddleware: Account=devstoreaccount1 Container=adonis-driver-test Blob=hello-moved.txt
2020-06-29T14:57:53.995Z 4d778175-3e41-48d0-9b27-ee2e8d2dbeee verbose: DispatchMiddleware: Dispatching request...
[...]

PR #469, opened 2020-07-08
[feature] Implementing syncCopyFromURL

AlexanderYW:
Hi, I have implemented syncCopyFromURL

I'm not quite sure about types for it, is there a place i can read about those?, or is the documentation for JS SDK valid in this case?
https://docs.microsoft.com/en-us/javascript/api/@azure/storage-blob/blobclient?view=azure-node-latest#synccopyfromurl-string--blobsynccopyfromurloptions-
If so whats the reason for startCopyFromURL having `tier` but syncCopyFromURL does not? (or aleast I can't find it)

Also prettier ran on `models.ts` on save, replacing ' with " as used in other files :)


AlexanderYW:
Issue: #461

XiaoningLiu:
Thanks @AlexanderYW for your great contribution! I'm OOF these days will take detailed look at next Monday,

AlexanderYW:
@XiaoningLiu did you get the time to look at my PR :)?

Would be nice if this feature could be a part of Azurite soon, since I want to use it in my own repo to test my driver for AdonisJs :)

src/blob/generated/artifacts/models.ts
```diff
@@ -154,6 +154,7 @@ export interface BlobProperties {
    * Possible values include: 'pending', 'success', 'aborted', 'failed'
    */
   copyStatus?: CopyStatusType;
+  syncCopyStatus?: SyncCopyStatusType;
```
XiaoningLiu, line 157: >syncCopyStatus?: SyncCopyStatusType; [](start = 2, length = 36)

This file (all files under generated folder) should be generated. 

You can extend generated mode in IBlobMetadataStore.ts if generated model doesn't fit our needs.

XiaoningLiu: Can we leverage existing `copyStatus` member?

---
In reply to: [457872339](https://github.com/Azure/Azurite/pull/469#discussion_r457872339) [](ancestors = 457872339)

src/blob/generated/artifacts/models.ts
```diff
@@ -5013,7 +5014,7 @@ export interface BlobCopyFromURLHeaders {
   /**
    * State of the copy operation identified by x-ms-copy-id. Possible values include: 'success'
    */
-  copyStatus?: SyncCopyStatusType;
+  copyStatus?: CopyStatusType;
```
XiaoningLiu, line 5017: >CopyStatusType [](start = 15, length = 14)

Why this change?

src/blob/handlers/BlobHandler.ts
```diff
@@ -721,6 +713,78 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
+    options: Models.BlobCopyFromURLOptionalParams,
+    context: Context
+  ): Promise<Models.BlobCopyFromURLResponse> {
+    const blobCtx = new BlobStorageContext(context);
+    const account = blobCtx.account!;
+    const container = blobCtx.container!;
+    const blob = blobCtx.blob!;
+
+    // TODO: Check dest Lease status, and set to available if it's expired, see sample in BlobHandler.setMetadata()
+    const url = new URL(copySource);
```
XiaoningLiu, line 736: BlobMetadataStore has implemented this in `new BlobWriteLeaseSyncer(doc).sync(lease);` .  Can do same thing in metatadata store implementation.

blueww:
This PR generally looks good to me.
It seems the blob copy will always  finish in the same request. This is true if inside same account, but if cross account copy, the copy blob request might won't finish in the copyblob request, but need check by get blob copy status later. So Azurite behavior might be different than server on copy big blob across account.

src/blob/persistence/LokiBlobMetadataStore.ts
```diff
@@ -1935,6 +1935,181 @@ export default class LokiBlobMetadataStore
...
+    );
+
+    validateWriteConditions(
+      context,
+      options.modifiedAccessConditions,
+      destBlob
+    );
+
+    if (destBlob) {
+      new BlobWriteLeaseValidator(options.leaseAccessConditions).validate(
```
XiaoningLiu, line 2002: >BlobWriteLeaseValidator [](start = 10, length = 23)

set lease state to available if it's expired by. 

```javascript
lease = new BlobLeaseAdapter(destBlob);
new BlobWriteLeaseSyncer(destBlob).sync(lease);
```

src/blob/authentication/OperationBlobSASPermission.ts
```diff
@@ -420,6 +427,13 @@ OPERATION_BLOB_SAS_CONTAINER_PERMISSIONS.set(
   Operation.Blob_AbortCopyFromURL,
   new OperationBlobSASPermission(BlobSASPermission.Write)
 );
+OPERATION_BLOB_SAS_CONTAINER_PERMISSIONS.set(
+  Operation.Blob_CopyFromURL,
+  // TODO: If destination is an existing blob, create permission is not enough
+  new OperationBlobSASPermission(
```
XiaoningLiu, line 433: We don't need this TODO comment since AccountSASAuthenticator and BlobSASAuthenticator already check this.

XiaoningLiu:
> This PR generally looks good to me.
> It seems the blob copy will always finish in the same request. This is true if inside same account, but if cross account copy, the copy blob request might won't finish in the copyblob request, but need check by get blob copy status later. So Azurite behavior might be different than server on copy big blob across account.

I think current implementation doesn't support cross account blob copy, which is fine for now.

XiaoningLiu:
@AlexanderYW Sorry for late. General looks good with some comments, please take a look.

src/blob/generated/artifacts/models.ts
AlexanderYW: Previously I had issues with building `autorest.typescript.server`, "solution" was to downgrade to nodejs v11.15.0 since the project is using gulp ^3.9.1

Now i only get this type error when generating
image

Beside that it generates the file now

src/blob/generated/artifacts/models.ts
AlexanderYW: To be honest i don't remember why I did that... but it seems to be working with `SyncCopyStatusType` as well :)

src/blob/persistence/LokiBlobMetadataStore.ts
AlexanderYW: Sorry i'm new to the codebase, and I'm not quite sure I understand what you mean

You wan't me to replease the WriteLeaseValidator with your suggestion or?

src/blob/authentication/OperationBlobSASPermission.ts
AlexanderYW: I have removed this TODO in my latest commit, should I remove the one from `Blob_StartCopyFromURL` as well then? :)

XiaoningLiu: Yes, feel free to remove all of them.

src/blob/persistence/LokiBlobMetadataStore.ts
XiaoningLiu: This is for one of your TODO comment in Handler implementation, which mentions need to set lease to avaiable when blob lease is expiry. So the recommendation is to add (instead of removing any) above 2 lines for that.

src/blob/generated/artifacts/models.ts
XiaoningLiu: You don't need to do a regeneration if this PR doesn't modify any files under generated foler. So I'm thinking if we can keep generated/artifacts/models.ts as it is. Any manual updates to files under generated folder will be override by next code generation.

What's the main purpose for adding `syncCopyStatus` property? If we can reuse `copyStatus` then don't need update the generated/artifacts/models.ts.

Or if it is a must to add `syncCopyStatus`, we can doing this by interface extend (see many other extended models in IBlobMetadataStore.ts), without manually updating generated/artifacts/models.ts. For example

```javascript
interface ExtendBlobProperties extends BlobProperties {
  syncCopyStatus?: SyncCopyStatusType
}
```

XiaoningLiu:
Hi @AlexanderYW, master branch has updated to use Azure Storage track2 SDK v12 for test cases. APIs in v12 is bit different with v10. Please check if any updates needed for test cases in this PR.

AlexanderYW:
Hi @XiaoningLiu, from what i can see there is no merge conflicts with the master branch and my changes, will check if any updates is needed for test cases later today if i get the time :)

src/blob/persistence/LokiBlobMetadataStore.ts
AlexanderYW: Hi @XiaoningLiu, I'm afraid I don't dont fully understand what they're doing, but I have tried implementing them

XiaoningLiu:
>     const result_startcopy = await destBlobClient.beginCopyFromURL(

Please keep useing `syncCopyFromURL` .

`beginCopyFromURL` is async copy

---
Refers to: tests/blob/apis/blob.test.ts:786 in c68d8c9. [](commit_id = c68d8c96c33a6ae61f4a4de40923743faf5426c8, deletion_comment = False)

AlexanderYW:
@XiaoningLiu when is this expected to be released :)?

XiaoningLiu:
Hi @AlexanderYW We have a planned release in incoming several weeks according to backend service deployment progress. Promisingly around end of Sep.

AlexanderYW:
Thanks for the update @XiaoningLiu :)

PR #524, opened 2020-07-28
DeleteTable Implementation

edwin-huber:
PR for #485 

Delete Table & Tests Implemented.

### DIFF
--- a/src/table/persistence/LokiTableMetadataStore.ts
+++ b/src/table/persistence/LokiTableMetadataStore.ts
@@ -103,8 +103,6 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
     if (tableColl != null) {
       this.db.removeCollection(name);
     } else {
-      // tslint:disable-next-line: no-console
-      console.log("DID NOT FIND:" + name);
       throw StorageErrorFactory.getTableNotFound(context);
     }
     // remove table reference from collection registry
--- a/src/table/persistence/LokiTableMetadataStore.ts
+++ b/src/table/persistence/LokiTableMetadataStore.ts
@@ -30,28 +30,30 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
   }
 
   public async createTable(context: Context, table: TableModel): Promise<void> {
-    const tablename = this.getTableCollectionNameString(
-      table.account,
-      table.tableName
-    );
+    // Check for table entry in the table registry collection
     const coll = this.db.getCollection(this.TABLE_COLLECTION);
     const doc = coll.findOne({
-      // ToDO REMOVE: accountName: table.account,
-      tableName: tablename
+      account: table.account,
+      name: table.tableName
     });
 
     // If the metadata exists, we will throw getTableAlreadyExists error
     if (doc) {
       throw StorageErrorFactory.getTableAlreadyExists(context);
     }
-    table.tableName = tablename;
     coll.insert(table);
-    const extentColl = this.db.getCollection(tablename);
+
+    // now we create the collection to represent the table using a unique string
+    const uniqueTableName = this.getUniqueTableCollectionName(
+      table.account,
+      table.tableName
+    );
+    const extentColl = this.db.getCollection(uniqueTableName);
     if (extentColl) {
       throw StorageErrorFactory.getTableAlreadyExists(context);
     }
 
-    this.db.addCollection(tablename, {
+    this.db.addCollection(uniqueTableName, {
       // Optimization for indexing and searching
       // https://rawgit.com/techfort/LokiJS/master/jsdoc/tutorial-Indexing%20and%20Query%20performance.html
       indices: ["ParititionKey", "RowKey"]
@@ -65,7 +67,7 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
     entity: IEntity
   ): Promise<void> {
     const tableColl = this.db.getCollection(
-      this.getTableCollectionNameString(account, tableName)
+      this.getUniqueTableCollectionName(account, tableName)
     );
     if (!tableColl) {
       throw StorageErrorFactory.getTableNotExist(context);
@@ -94,20 +96,24 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
   public async deleteTable(
     context: Context,
     name: string,
-    account: string
+    accountName: string
   ): Promise<void> {
-    const tablename = this.getTableCollectionNameString(account, name);
-    const tableColl = this.db.getCollection(tablename);
+    const uniqueTableName = this.getUniqueTableCollectionName(
+      accountName,
+      name
+    );
+    const tableColl = this.db.getCollection(uniqueTableName);
     // delete the collection / table
     if (tableColl != null) {
-      this.db.removeCollection(tablename);
+      this.db.removeCollection(uniqueTableName);
     } else {
       throw StorageErrorFactory.getTableNotFound(context);
     }
     // remove table reference from collection registry
     const coll = this.db.getCollection(this.TABLE_COLLECTION);
     const doc = coll.findOne({
-      tableName: tablename
+      account: accountName,
+      tableName: name
     });
     if (doc != null) {
       coll.remove(doc);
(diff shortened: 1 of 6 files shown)


src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -29,29 +29,29 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
...
+      table.account,
+      table.tableName
+    );
     const coll = this.db.getCollection(this.TABLE_COLLECTION);
     const doc = coll.findOne({
-      accountName: table.account,
-      name: table.tableName
+      // ToDO REMOVE: accountName: table.account,
+      tableName: tablename
     });
```
XiaoningLiu, line 41: Any specific reason for this change?

Here we'd better use `accountName` and original `tableName` as entity index inside TABLE_COLLECTION.

Reason for this it's helpful to implement QueryTables, we can return entites from TABLE_COLLECTION directly instead of doing another string split to get original table name from combined accountname_tablename

src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -215,4 +235,10 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
 
     this.closed = true;
   }
+  private getTableCollectionNameString(
+    accountName: string,
+    tableName: string
+  ): string {
+    return accountName + tableName;
```
XiaoningLiu, line 242: >accountName + tableName; [](start = 11, length = 24)

We need a character between accountName and tableName.

For example, add an additional '$', like 
return `${accountName}$${tableName}`. 

This is used to differentiate following 2 race condition scenarios:

#1 accountName = "abc", tableName="efg". Then we will get "abcefg"

#2 accountName= "ab", tableName= "cefg". Then we still get "abcefg"

src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -215,4 +235,10 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
 
     this.closed = true;
   }
+  private getTableCollectionNameString(
```
XiaoningLiu, line 238: >getTableCollectionNameString [](start = 10, length = 28)

How about rename to `getTableCollectionName` without "string" as appendix. 

Different with JavaScript, TypeScript method declaration shows return type.

XiaoningLiu, line 238: >private [](start = 2, length = 7)

Needs an additional line between 2 methods definitions.

tests/table/apis/table.test.ts
```diff
@@ -5,7 +5,13 @@ import { configLogger } from "../../../src/common/Logger";
...
 import TableServer from "../../../src/table/TableServer";
 import { TABLE_API_VERSION } from "../../../src/table/utils/constants";
-import { EMULATOR_ACCOUNT_KEY, EMULATOR_ACCOUNT_NAME, getUniqueName, overrideRequest } from "../../testutils";
+import {
+  EMULATOR_ACCOUNT_KEY,
+  EMULATOR_ACCOUNT_NAME,
+  getUniqueName,
+  overrideRequest
+} from "../../testutils";
+// import { options } from "args";
```
XiaoningLiu, line 14: >options [](start = 12, length = 7)

Remove commented code here.

src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -29,29 +29,29 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
     return this.closed;
   }
 
-  public async createTable(
-    context: Context,
-    table: TableModel
-  ): Promise<void> {
+  public async createTable(context: Context, table: TableModel): Promise<void> {
+    const tablename = this.getTableCollectionNameString(
```
XiaoningLiu, line 33: >tablename [](start = 10, length = 9)

Better use `tableCollectionName`. It's bit confusing because it's not the original tableName.

src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -29,29 +29,29 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
...
+      // ToDO REMOVE: accountName: table.account,
+      tableName: tablename
     });
 
     // If the metadata exists, we will throw getTableAlreadyExists error
     if (doc) {
       throw StorageErrorFactory.getTableAlreadyExists(context);
     }
-
+    table.tableName = tablename;
```
XiaoningLiu, line 47: >table.tableName = tablename; [](start = 4, length = 28)

I think you are assigning tableCollectionName to table.tableName.

It's better keep tableName refer to the original table name. Add another separate property tableCollectionName into TableModel for the table collection name.

src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -90,10 +93,27 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
 
   public async deleteTable(
     context: Context,
-    tableName: string
+    name: string,
+    account: string
   ): Promise<void> {
-    // TODO    context: Context
-    throw new NotImplementedError();
+    const tablename = this.getTableCollectionNameString(account, name);
```
XiaoningLiu, line 99: >tablename [](start = 10, length = 9)

Better use `tableCollectionName` instead of `tableName`

src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -90,10 +93,27 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
...
+    // delete the collection / table
+    if (tableColl != null) {
+      this.db.removeCollection(tablename);
+    } else {
+      throw StorageErrorFactory.getTableNotFound(context);
+    }
+    // remove table reference from collection registry
+    const coll = this.db.getCollection(this.TABLE_COLLECTION);
+    const doc = coll.findOne({
+      tableName: tablename
```
XiaoningLiu, line 110: >tableName: tablename [](start = 6, length = 20)

TABLE_COLLECTION has `accountName` and `tableName` as index. Here `tableName` refers to the orignal table name instead of tableCollectionName.

So should use 
```javascript
const doc = coll.findOne({
  accountName: account,
  tableName: name
});
```

Or after adding `tableCollectionName` into TABLE_COLLECTION index. We can search with following. But seems we don't need tableCollectionName as collection index because indexes `accountName` and `tableName`  are enough for use.

```javascript
const doc = coll.findOne({
  tableCollectionName
});
```

tests/table/apis/table.test.ts
```diff
@@ -120,4 +126,49 @@ describe("table APIs test", () => {
...
+
+  it("deleteTable that exists, @loki", done => {
+    /*
+    https://docs.microsoft.com/en-us/rest/api/storageservices/delete-table
+    */
+    requestOverride.headers = {
+      Prefer: "return-content",
+      accept: "application/json;odata=fullmetadata"
+    };
+
```
XiaoningLiu, line 138: Delete table API document doesn't include accept and Prefer headers. So we can use SDK default values by assign {} to requestOverride.headers.

requestOverride.headers = {}
      Prefer: "return-content",
      accept: "application/json;odata=fullmetadata"
    };}

tests/table/apis/table.test.ts
```diff
@@ -120,4 +126,49 @@ describe("table APIs test", () => {
...
+    });
+  });
+
+  it("deleteTable that does not exist, @loki", done => {
+    /*
+    https://docs.microsoft.com/en-us/rest/api/storageservices/delete-table
+    */
+    requestOverride.headers = {
+      accept: "application/json;"
+    };
```
XiaoningLiu, line 165: Delete table API document doesn't include accept and Prefer headers. So we can use SDK default values by assign {} to requestOverride.headers.

requestOverride.headers = {}
      Prefer: "return-content",
      accept: "application/json;odata=fullmetadata"
    };}

tests/table/apis/table.test.ts
```diff
@@ -120,4 +126,49 @@ describe("table APIs test", () => {
...
+
+    const tableToDelete = tableName + "causeerror";
+
+    tableService.deleteTable(tableToDelete, (error, result) => {
+      assert.equal(result.statusCode, 404); // no body expected, we expect 404
```
XiaoningLiu, line 170: >assert [](start = 6, length = 6)

Can we further check storage error code string? Like `TableNotFound`

src/table/persistence/LokiTableMetadataStore.ts
edwin-huber: now using uniqueTableName

PR #525, opened 2020-07-29
Table delete entity

edwin-huber:
I'm getting a ECONNRESET error when running the entity tests that I have written to test the etag matching error...
I can't explain it. I'll submit the PR now so that you can take a look.  

I'm hoping fresh eyes will spot the error in the morning! :smile:

### DIFF
--- a/src/table/handlers/TableHandler.ts
+++ b/src/table/handlers/TableHandler.ts
@@ -303,7 +303,32 @@ export default class TableHandler extends BaseHandler implements ITableHandler {
     //   version: "version"
     // };
     // TODO
-    throw new NotImplementedError();
+
+    const tableCtx = new TableStorageContext(context);
+    const accountName = tableCtx.account;
+    const partitionKey = tableCtx.partitionKey!; // Get partitionKey from context
+    const rowKey = tableCtx.rowKey!; // Get rowKey from context
+
+    if (!partitionKey || !rowKey) {
+      throw StorageErrorFactory.getPropertiesNeedValue(context);
+    }
+
+    // currently the props are not coming through as args, so we take them from the table context
+    await this.metadataStore.deleteTableEntity(
+      context,
+      tableCtx.tableName!,
+      accountName!,
+      partitionKey,
+      rowKey
+    );
+
+    return {
+      statusCode: 204,
+      date: tableCtx.startTime,
+      clientRequestId: options.requestId,
+      requestId: tableCtx.contextID,
+      version: TABLE_API_VERSION
+    };
   }
 
   public async insertEntity(
--- a/src/table/handlers/TableHandler.ts
+++ b/src/table/handlers/TableHandler.ts
@@ -289,37 +289,28 @@ export default class TableHandler extends BaseHandler implements ITableHandler {
     options: Models.TableDeleteEntityOptionalParams,
     context: Context
   ): Promise<Models.TableDeleteEntityResponse> {
-    // e.g
-    // const tableCtx = new TableStorageContext(context);
-    // const accountName = tableCtx.account;
-    // const tableName = tableCtx.tableName; // Get tableName from context
-    // const partitionKey = tableCtx.partitionKey!; // Get partitionKey from context
-    // const rowKey = tableCtx.rowKey!; // Get rowKey from context
-    // return {
-    //   statusCode: 204,
-    //   date: tableCtx.startTime,
-    //   clientRequestId: "clientRequestId",
-    //   requestId: "requestId",
-    //   version: "version"
-    // };
-    // TODO
-
     const tableCtx = new TableStorageContext(context);
     const accountName = tableCtx.account;
     const partitionKey = tableCtx.partitionKey!; // Get partitionKey from context
     const rowKey = tableCtx.rowKey!; // Get rowKey from context
-
+    let etag: string = "";
+    if (tableCtx.request !== undefined) {
+      etag = tableCtx.request.getHeader("match") as string;
+    }
     if (!partitionKey || !rowKey) {
       throw StorageErrorFactory.getPropertiesNeedValue(context);
     }
-
+    if (etag === "" || etag === undefined) {
+      throw StorageErrorFactory.getPreconditionFailed(context);
+    }
     // currently the props are not coming through as args, so we take them from the table context
     await this.metadataStore.deleteTableEntity(
       context,
       tableCtx.tableName!,
       accountName!,
       partitionKey,
-      rowKey
+      rowKey,
+      etag!
     );
 
     return {
--- a/src/table/handlers/TableHandler.ts
+++ b/src/table/handlers/TableHandler.ts
@@ -293,14 +293,11 @@ export default class TableHandler extends BaseHandler implements ITableHandler {
     const accountName = tableCtx.account;
     const partitionKey = tableCtx.partitionKey!; // Get partitionKey from context
     const rowKey = tableCtx.rowKey!; // Get rowKey from context
-    let etag: string = "";
-    if (tableCtx.request !== undefined) {
-      etag = tableCtx.request.getHeader("match") as string;
-    }
+
     if (!partitionKey || !rowKey) {
       throw StorageErrorFactory.getPropertiesNeedValue(context);
     }
-    if (etag === "" || etag === undefined) {
+    if (ifMatch === "" || ifMatch === undefined) {
       throw StorageErrorFactory.getPreconditionFailed(context);
     }
     // currently the props are not coming through as args, so we take them from the table context
@@ -310,7 +307,7 @@ export default class TableHandler extends BaseHandler implements ITableHandler {
       accountName!,
       partitionKey,
       rowKey,
-      etag!
+      ifMatch
     );
 
     return {
--- a/tests/table/apis/table.entity.test.ts
+++ b/tests/table/apis/table.entity.test.ts
@@ -4,7 +4,6 @@ import * as Azure from "azure-storage";
 import { configLogger } from "../../../src/common/Logger";
 import TableConfiguration from "../../../src/table/TableConfiguration";
 import TableServer from "../../../src/table/TableServer";
-// import { TABLE_API_VERSION } from "../../../src/table/utils/constants";
 import {
   EMULATOR_ACCOUNT_KEY,
   EMULATOR_ACCOUNT_NAME,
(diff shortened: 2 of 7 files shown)


blueww:
@edwin-huber 
Thanks for raise the PR!
See the following case failed on several platforms. Would you please make the test pass? 
 

>  1) table Entity APIs test
>        Should not delete an Entity not matching Etag, @loki:
>      Uncaught AssertionError [ERR_ASSERTION]: ifError got unwanted exception: The update condition specified in the request was not satisfied.
> RequestId:18cc480a-4566-48e7-99cd-7eec1b680263
> Time:2020-07-29T18:53:14.457Z
>       at Function.StorageServiceClient._normalizeError (node_modules\azure-storage\lib\common\services\storageserviceclient.js:1205:23)
>       at TableService.StorageServiceClient._processResponse (node_modules\azure-storage\lib\common\services\storageserviceclient.js:751:50)
>       at Request.processResponseCallback [as _callback] (node_modules\azure-storage\lib\common\services\storageserviceclient.js:319:37)
>       at Request.self.callback (node_modules\request\request.js:185:22)
>       at Request.<anonymous> (node_modules\request\request.js:1161:10)
>       at IncomingMessage.<anonymous> (node_modules\request\request.js:1083:12)
>       at endReadableNT (_stream_readable.js:1221:12)
>       at processTicksAndRejections (internal/process/task_queues.js:84:21)
>

src/table/handlers/TableHandler.ts
```diff
@@ -289,21 +289,37 @@ export default class TableHandler extends BaseHandler implements ITableHandler {
...
-    // TODO
-    throw new NotImplementedError();
+    const tableCtx = new TableStorageContext(context);
+    const accountName = tableCtx.account;
+    const partitionKey = tableCtx.partitionKey!; // Get partitionKey from context
+    const rowKey = tableCtx.rowKey!; // Get rowKey from context
+    let etag: string = "";
+    if (tableCtx.request !== undefined) {
+      etag = tableCtx.request.getHeader("match") as string;
+    }
```
XiaoningLiu, line 299: Get etag from parameter `ifMatch`. Don't need get from context request.

tests/table/apis/table.entity.test.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    /* https://docs.microsoft.com/en-us/rest/api/storageservices/delete-entity1#request-headers
+    If-Match	Required. The client may specify the ETag for the entity on the request in
+    order to compare to the ETag maintained by the service for the purpose of optimistic concurrency.
+    The delete operation will be performed only if the ETag sent by the client matches the value
+    maintained by the server, indicating that the entity has not been modified since it was retrieved by the client.
+    To force an unconditional delete, set If-Match to the wildcard character (*). */
+    requestOverride.headers = {
+      Match: "*"
+    };
```
XiaoningLiu, line 97: Don't need to override header here, as table SDK supports if-match.

Use following way to specific if-match for delete entity.


```javascript
const entity = {
    PartitionKey: "part1",
    RowKey: "row1",
    myValue: "somevalue",
  ".metadata": {
    etag: "*"
  }
}

tableService.deleteEntity(entity)
```

tests/table/apis/table.entity.test.ts
```diff
@@ -0,0 +1,188 @@
...
+      RowKey: "row1",
+      myValue: "value1"
+    };
+    tableService.insertEntity(tableName, entity, (error, result, response) => {
+      assert.equal(response.statusCode, 201);
+      done();
+    });
+  });
+
+  it("Should delete an Entity given partition key and row key in the URL, @loki", done => {
```
XiaoningLiu, line 81: >Should delete an Entity given partition key and row key in the URL, @loki [](start = 6, length = 73)

Each test case should be able to run independently, instead of requiring other test cases. This is helpful to run and debug single test case without re-run all test cases.

For example, to test delete entity, this test case should insert entity first, then delete entity.

tests/table/apis/table.entity.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    };
+    tableService.insertEntity(
+      tableName,
+      entity,
+      (inserterror, insertresult, insertresponse) => {
+        if (!inserterror) {
+          sleep(10);
+          requestOverride.headers = {
+            Match: "0x2252C97588D4000" // just a random etag
+          };
```
XiaoningLiu, line 124: Use ".metadata" property to customize if-match etag for an entity to be deleted.

```
requestOverride.headers = {};
entity[".metadata"]["etag"] = "0x2252C97588D4000"; // just a random etag
```

tests/table/apis/table.entity.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    requestOverride.headers = {
+      Prefer: "return-content",
+      accept: "application/json;odata=fullmetadata"
+    };
+    tableService.insertEntity(
+      tableName,
+      entity,
+      (inserterror, insertresult, insertresponse) => {
+        if (!inserterror) {
+          sleep(10);
```
XiaoningLiu, line 121: >sleep(10); [](start = 10, length = 10)

Why sleep 10ms here? Ideally we don't need to sleep.

tests/table/apis/table.entity.test.ts
```diff
@@ -0,0 +1,188 @@
...
+      myValue: "shouldnotmatchetag"
+    };
+    requestOverride.headers = {
+      Prefer: "return-content",
+      accept: "application/json;odata=fullmetadata"
+    };
+    tableService.insertEntity(
+      tableName,
+      entity,
+      (inserterror, insertresult, insertresponse) => {
```
XiaoningLiu, line 119: >inserterror [](start = 7, length = 11)

Azurite uses camel-case naming, it's better to use `insertError`, `insertResult` and `insertResponse`.

tests/table/apis/table.entity.test.ts
```diff
@@ -0,0 +1,188 @@
...
+        if (!inserterror) {
+          sleep(10);
+          requestOverride.headers = {
+            Match: "0x2252C97588D4000" // just a random etag
+          };
+          // this delete should fail as we modified the etag, but we get ECONRESET???
```
XiaoningLiu, line 125: >but we get ECONRESET??? [](start = 62, length = 23)

I don't get ECONRESET, did you trigger any breaking point in Azurite handler?

tests/table/apis/table.entity.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    };
+    tableService.insertEntity(
+      tableName,
+      entity,
+      (error, result, insertresponse) => {
+        if (!error) {
+          const etag: string = result[".metadata"].etag;
+          requestOverride.headers = {
+            Match: etag
+          };
```
XiaoningLiu, line 166: Same with above, you can just pass `result` into deleteEntity. `result` is a SDK defined entity type.

```
if (!error) {
          tableService.deleteEntity(
            tableName,
            result,
```

tests/table/apis/table.entity.test.ts
```diff
@@ -0,0 +1,188 @@
...
+          // this delete should fail as we modify the etag
```
XiaoningLiu, line 167: >// this delete should fail as we modify the etag [](start = 10, length = 48)

Following assert expect delete to be successful, but description here wants the delete to be failed.

tests/table/apis/table.entity.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    // https://docs.microsoft.com/en-us/rest/api/storageservices/delete-entity1
+    const entity = {
+      PartitionKey: "part1",
+      RowKey: "row3",
+      myValue: "shouldmatchetag"
+    };
+    requestOverride.headers = {
+      Prefer: "return-content",
+      accept: "application/json;odata=fullmetadata"
+    };
```
XiaoningLiu, line 157: Don't need override headers, as Delete Entity doesn't support prefer and accept. [https://docs.microsoft.com/en-us/rest/api/storageservices/delete-entity1](https://docs.microsoft.com/en-us/rest/api/storageservices/delete-entity1)

tests/table/apis/table.entity.test.ts
```diff
@@ -0,0 +1,188 @@
...
+          );
+        } else {
+          assert.ifError(inserterror);
+          done();
+        }
+      }
+    );
+  });
+
+  it("Should not delete an matching Etag, @loki", done => {
```
XiaoningLiu, line 147: >Should not delete an matching Etag [](start = 6, length = 34)

Following assert expect delete to be successful, but description here wants the delete to be failed.

tests/table/apis/table.entity.test.ts
```diff
@@ -0,0 +1,188 @@
...
+  it("Should not delete an Entity not matching Etag, @loki", done => {
+    // https://docs.microsoft.com/en-us/rest/api/storageservices/delete-entity1
+    const entity = {
+      PartitionKey: "part1",
+      RowKey: "row2",
+      myValue: "shouldnotmatchetag"
+    };
+    requestOverride.headers = {
+      Prefer: "return-content",
+      accept: "application/json;odata=fullmetadata"
```
XiaoningLiu, line 114: Delete entity doesn't support following 2 headers: prefer and accept. So we can reset headers override requestOverride.headers = {}

tests/table/apis/table.entity.test.ts
edwin-huber: I was trying to see if the TCP connection reset is due to a race condition somehow.  
I have removed the sleep.  
It seems that the tests fail in the case that there was an unsuccessful test of the test
```Should not delete an Entity not matching Etag, @loki``` in a previous run.  
I think we need to make the tests more idempotent, and create a new DB for each test run.

tests/table/apis/table.entity.test.ts
edwin-huber: Seems to be working fine now...

tests/table/apis/table.entity.test.ts
edwin-huber: Header override is for "insert"

tests/table/apis/table.entity.test.ts
edwin-huber: With all tests passing, this no longer causes issues for subsequent runs

PR #528, opened 2020-08-02
Add update Entity feature

sonniiy:
Added the update Entity feature and tests

### DIFF
--- a/src/table/persistence/LokiTableMetadataStore.ts
+++ b/src/table/persistence/LokiTableMetadataStore.ts
@@ -143,12 +143,31 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
 
   public async updateTableEntity(
     context: Context,
-    table: string,
-    partitionKey: string,
-    rowKey: string
+    tableName: string,
+    account: string,
+    entity: IEntity
   ): Promise<void> {
-    // TODO
-    throw new NotImplementedError();
+    const tableColl = this.db.getCollection(
+      this.getUniqueTableCollectionName(account, tableName)
+    );
+
+    // Throw errow, when table not exists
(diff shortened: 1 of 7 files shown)


src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -143,12 +143,38 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
...
+      throw StorageErrorFactory.getEntityNotExist(context);
+    } else {
+      // Test if etag value is valid
+      if (etag !== "*" && currentDoc.eTag !== etag) {
+        throw StorageErrorFactory.getPreconditionFailed(context);
+      }
+    }
+
+    tableColl.insert(entity);
+    tableColl.remove(currentDoc);
```
XiaoningLiu, line 177: Curious why there is no conflict error pops up. Because here insert new entity before removing existing. Both of them have same PartitionKey and RowKey.

XiaoningLiu:
>       indices: ["ParititionKey", "RowKey"]

Should be a typo, can you help update it to `ParititionKey`?

---
Refers to: src/table/persistence/LokiTableMetadataStore.ts:59 in 3221cb0. [](commit_id = 3221cb0a22e69e3b86980f7558a2ca0ac3f62da4, deletion_comment = False)

XiaoningLiu:
General looks good! Minor comments provided.

src/table/persistence/LokiTableMetadataStore.ts
sonniiy: I have changed the order

XiaoningLiu:
Take another run

PR #535, opened 2020-08-15
enable cross-account async copy

### LINKED ISSUE #530: [blob]: enable cross account async copy (same server)

Blob 

3.8.0

DockerHub

copying a blob which exists in a different storage account does not work and instead returns a 404. it turns out this is due to https://github.com/Azure/Azurite/blob/master/src/blob/handlers/BlobHandler.ts#L627, which explicitly returns `blobNotFound` in the case that the source storage account isn't the same as the destination storage account. as it appeared to handle this fine all the way down the stack (at least in the Loki provider), i removed the comparison and built a new local image. upon rerunning our local tests this worked fine, so this is more of a question.

is cross-account blob copy explicitly disabled? it appears to be handled in the backend. 

- setup 2 storage accounts
- create a blob on account 1
- attempt to copy the blob to account 2 using account 1 as the copy source

locally i was able to just remove the offending line as described above. any reason this couldn't be included in master?

patrickcarnahan:
Fixes #530.

enables cross-account copy for `startCopyFromURL` if the accounts both exist on the same `azurite` instance. this is accomplished by issuing a `GET` request to the provided `copySource` URL, which would need to include a SAS token in order for cross-account copy to function properly. if the response is `OK`, then the validation succeeds and the current metadata copy (no content copy) logic is used.


src/blob/handlers/BlobHandler.ts
```diff
@@ -624,14 +624,48 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
+
+        throw StorageErrorFactory.getBlobNotFound(context.contextId!);
+      }
+
+      this.logger.debug(
+        `BlobHandler:startCopyFromURL() Validating access to the source account ${sourceAccount}`,
+        context.contextId
+      );
+
+      const validationRequest = new fetch.Request(copySource, { method: "HEAD" });
```
patrickcarnahan, line 652: @XiaoningLiu i'm not super familiar with typescript/nodejs, but after some digging found `node-fetch` as a good http request library which implements `async/await` patterns. if you had something else in mind, such as `axios` (also available currently in the `node_modules`) then i'm happy to alter the implementation.

src/blob/utils/constants.ts
```diff
@@ -21,8 +21,17 @@ export const LOGGER_CONFIGS = {};
...
+export const EMULATOR_ACCOUNT_KEY_STR =
+  "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==";
 export const EMULATOR_ACCOUNT_KEY = Buffer.from(
-  "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==",
+  EMULATOR_ACCOUNT_KEY_STR,
+  "base64"
+);
+export const EMULATOR_ACCOUNT2_NAME = "devstoreaccount2";
+export const EMULATOR_ACCOUNT2_KEY_STR =
+  "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==";
```
patrickcarnahan, line 32: this should be a different value, wasn't sure how the first one was generated.

blueww: This is a base64 encoded string. 
You might can just  base64 encode a string and use it as key.

blueww:
Please fix the CI failure.

src/blob/utils/constants.ts
patrickcarnahan: @blueww thanks! i ended up just doing this:

```
PS > [Guid]::NewGuid().ToByteArray() | base64
```

let me know if you had something else in mind.

src/blob/handlers/BlobHandler.ts
```diff
@@ -624,14 +625,50 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
-      sourceAccount !== blobCtx.account ||
       sourceAccount === undefined ||
       sourceContainer === undefined ||
       sourceBlob === undefined
     ) {
       throw StorageErrorFactory.getBlobNotFound(context.contextId!);
     }
 
+    if (sourceAccount !== blobCtx.account) {
+      // Currently the only cross-account copy support is from/to the same server. In either case access is determined
```
XiaoningLiu, line 636:
```suggestion
      // Currently the only cross-account copy support is from/to the same Azurite instance. In either case access is determined
```

src/blob/handlers/BlobHandler.ts
```diff
@@ -624,14 +625,50 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
+      // by performing a request to the copy source to see if the authentication is valid.
+      const currentServer = blobCtx.request!.getHeader("Host") || "";
+      if (currentServer !== url.host) {
+        this.logger.error(
+          `BlobHandler:startCopyFromURL() Source account ${url} is not on the same server as target account ${account}`,
```
XiaoningLiu, line 641:
```suggestion
          `BlobHandler:startCopyFromURL() Source account ${url} is not on the same Azurite instance as target account ${account}`,
```

src/blob/handlers/BlobHandler.ts
```diff
@@ -624,14 +625,50 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
+
+        throw StorageErrorFactory.getBlobNotFound(context.contextId!);
+      }
+
+      this.logger.debug(
+        `BlobHandler:startCopyFromURL() Validating access to the source account ${sourceAccount}`,
+        context.contextId
+      );
+
+      const validationRequest = new fetch.Request(copySource, {
```
XiaoningLiu, line 653: Here we want to send out a request to validate accessing to copy source. Please use Node.js built-in `http` module for the request instead of introduing a new dependency just for one request.

See https://nodejs.org/api/http.html#http_http_request_url_options_callback

blueww:
Please also update [Support Matrix in ReadMe](https://github.com/Azure/Azurite#support-matrix) and [Change log](https://github.com/Azure/Azurite/blob/master/ChangeLog.md)

package.json
```diff
@@ -28,6 +28,7 @@
     "morgan": "^1.9.1",
```
blueww, line 28: Please also update [Support Matrix in ReadMe](https://github.com/Azure/Azurite#support-matrix) and [Change log](https://github.com/Azure/Azurite/blob/master/ChangeLog.md)

src/blob/handlers/BlobHandler.ts
```diff
@@ -624,14 +625,50 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
+          `BlobHandler:startCopyFromURL() Successfully validated access to source account ${sourceAccount}`,
+          context.contextId
+        );
+      } else {
+        this.logger.debug(
+          `BlobHandler:startCopyFromURL() Access denied to source account ${sourceAccount}. StatusCode=${validationResponse.status}`,
+          context.contextId
+        );
+
+        throw StorageErrorFactory.getBlobNotFound(context.contextId!);
```
XiaoningLiu, line 668: [Minor]

Azure Storage blob copy will return exactly same status code returned by HEAD request to copy source when doing multi account copy. You can try on a real Azure Storage account and send out requests to blob on another storage account. You can see a 403 error returned by Copy Blob when copy source includes an invalid shared access signature.

It's better to define a new error type generator method `getCannotVerifyCopySource()` in StorageErrorFactory. And return same HTTP status code with the HEAD request response. The error response body should align with real Azure Storage too.

For example, if HEAD request returns 400 instead of 200. Then Copy Blob should throw getCannotVerifyCopySource() with 400 as HTTP status code and `CannotVerifyCopySource` as storage error code.

src/blob/handlers/BlobHandler.ts
```diff
@@ -624,14 +625,50 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
     if (
-      sourceAccount !== blobCtx.account ||
       sourceAccount === undefined ||
       sourceContainer === undefined ||
       sourceBlob === undefined
     ) {
       throw StorageErrorFactory.getBlobNotFound(context.contextId!);
     }
 
+    if (sourceAccount !== blobCtx.account) {
```
blueww, line 635: From cross account blob copy, need check source blob tier:
from [rest Spec](https://docs.microsoft.com/en-us/rest/api/storageservices/copy-blob)

- The copy source and destination should be the same storage account when the source is archived. The request will fail with Conflict if the source of the copy is still in pending rehydrate state.

src/blob/utils/constants.ts
```diff
@@ -21,8 +21,17 @@ export const LOGGER_CONFIGS = {};
...
 export const DEFAULT_WRITE_CONCURRENCY_PER_LOCATION = 50;
 export const EMULATOR_ACCOUNT_NAME = "devstoreaccount1";
+export const EMULATOR_ACCOUNT_KEY_STR =
+  "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==";
 export const EMULATOR_ACCOUNT_KEY = Buffer.from(
-  "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==",
+  EMULATOR_ACCOUNT_KEY_STR,
+  "base64"
+);
+export const EMULATOR_ACCOUNT2_NAME = "devstoreaccount2";
```
XiaoningLiu, line 30: Please don't add `devstoreaccount2` into Azurite as another default built-in account.

I think here you want to add another account just for testing purpose. You can adding another account by updating environment variable dynamically.

`process.env[AZURITE_ACCOUNTS] ="account1:key1:key2;account2:key1:key2"`

See environment variable format and requirements https://github.com/azure/azurite#customized-storage-accounts--keys.

Or you can import and customize `DEFAULT_EMULATOR_ACCOUNTS` in your test case file instead modifying it in AccountDataStore.ts directly.

src/blob/handlers/BlobHandler.ts
```diff
@@ -624,14 +625,50 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
+          `BlobHandler:startCopyFromURL() Access denied to source account ${sourceAccount}. StatusCode=${validationResponse.status}`,
+          context.contextId
+        );
+
+        throw StorageErrorFactory.getBlobNotFound(context.contextId!);
+      }
+    }
+
     // Preserve metadata key case
     const metadata = convertRawHeadersToMetadata(
```
XiaoningLiu, line 673: It seems good to leverage existing `metadataStore.startCopyFromURL` implementation which should in `LokiBlobMetadataStore.startCopyFromURL`. 

But please double check with if there any differences between single account copy with cross account copy. For example, if all request headers are appliable for cross account copy in https://docs.microsoft.com/en-us/rest/api/storageservices/copy-blob#request-headers

You can do a quick check with your real Azure Storage account.

tests/blob/sas.test.ts
```diff
@@ -1124,4 +1143,99 @@ describe("Shared Access Signature (SAS) authentication", () => {
 
     await containerClient.delete();
   });
+
+  it("Copy blob across accounts should require SAS token @loki", async () => {
```
blueww, line 1147: Please reference [rest API](https://docs.microsoft.com/en-us/rest/api/storageservices/copy-blob) to add more test cases, like:
1. copy fail when source blob tier is archive
2. when source container is public, copy from source blob Uri without SAS should work
3. If the copy request has "x-ms-meta-name:value" header, dest blob will has meta data as copy request; else dest will have same meta as source.
4.

src/blob/handlers/BlobHandler.ts
patrickcarnahan: @XiaoningLiu thanks! the reason i didn't pursue the built-in `http` module originally was due to the fact that it did not support `async/await`. i definitely like this approach better.

patrickcarnahan: @XiaoningLiu after some investigation it would appear that the built-in `http` library doesn't actually support being used with `promisify` since it uses non-standard callback patterns. as `ms-rest-js` already brings in `axios` I have switched to that library instead to avoid adding a new dependency.

src/blob/handlers/BlobHandler.ts
patrickcarnahan: I will certainly play around and see how I can best mimic the behavior of the real APIs.

package.json
patrickcarnahan: @blueww which version should I place the changes under? 3.8.0?

src/blob/handlers/BlobHandler.ts
patrickcarnahan: @XiaoningLiu I played around with blob here a bit and found it a tad difficult to exactly mimic the behavior that blob provides. I ended up issuing a `GET` request with `?comp=metadata` in lieu of the `HEAD` call since the former actually returns the error details in the response while the latter does not. Let me know if the changes I have provided are not what you had intended.

src/blob/utils/constants.ts
patrickcarnahan: Done! Thanks for the tip, this is my lack of `node` and `typescript` experience showing through. :)

tests/blob/sas.test.ts
patrickcarnahan: @blueww I have added tests for all but the source blob tier validation. I will be looking into that next.

src/blob/handlers/BlobHandler.ts
patrickcarnahan: Added a test which validates this behavior. It would appear that `Azurite` currently always raises a `409` if the source blob is archived regardless of any other conditions. As there is now a regression test in place for this behavior I request that any changes to the existing behavior be moved into a separate issue.

src/blob/handlers/BlobHandler.ts
patrickcarnahan: @XiaoningLiu I read through those docs and didn't notice anything that is specific to cross account vs. within the same account.

tests/blob/sas.test.ts
patrickcarnahan: Tests for all of the cases you mentioned above have been added.

src/blob/handlers/BlobHandler.ts
```diff
@@ -624,14 +627,87 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
+        } else {
+          // For non-successful responses attempt to unwrap the error message from the metadata call.
+          let message: string =
+            "Could not verify the copy source within the specified time.";
+          if (
+            validationResponse.headers[HeaderConstants.CONTENT_TYPE] ===
+            "application/xml"
+          ) {
+            const authenticationError = await parseXML(validationResponse.data);
+            if (authenticationError.Message !== undefined) {
```
patrickcarnahan, line 697: this chunk was in response to an observation of the azure api behavior. when there was an issue with the authentication (sas token) it wouldn't actually return `CannotVerifyCopySource` but instead was returning `AuthenticationFailed` as the code and the message described what was wrong with authentication. currently there is a discrepancy which could be fixed, which is to set the code and message entirely from the remote call instead of always using `CannotVerifyCopySource`.

the second piece of this that was odd is azurite returns `AuthorizationFailure` as the error code if, for instance, the SAS token is invalid. azure storage, however, returns `AuthenticationFailed` as the error code. is this something that you have noticed?

src/blob/handlers/BlobHandler.ts
```diff
@@ -624,14 +627,87 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
+              message = authenticationError.Message.replace(/\n+/gm, "");
```
patrickcarnahan, line 698: what i found was the error message returned from the caller could not then be re-encoded as the message using the `StorageErrorFactory` below. in order to work around the string issue i removed the `\n` characters from the message returned by the remote endpoint, allowing the remote endpoint message to be used.

src/blob/handlers/BlobHandler.ts
```diff
@@ -624,14 +627,87 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
...
+
+      this.logger.debug(
+        `BlobHandler:startCopyFromURL() Validating access to the source account ${sourceAccount}`,
+        context.contextId
+      );
+
+      // In order to retrieve proper error details we make a metadata request to the copy source. If we instead issue
+      // a HEAD request then the error details are not returned and reporting authentication failures to the caller
+      // becomes a black box.
+      const metadataUrl = URLBuilder.parse(copySource);
```
patrickcarnahan, line 662: i switched to using a `GET` instead of `HEAD` due to the authentication error details not getting returned. to ensure we do not download blob content i ensure the query contains `comp=metadata`.

package.json
blueww: @patrickcarnahan 
You might can add a tag as "## Upcoming Release", and we will update the release number later.

@XiaoningLiu 
Any concern if we add a tag "## Upcoming Release" in change log?

blueww: @patrickcarnahan
The change log looks good to me.

XiaoningLiu:
@ patrickcarnahan Thanks for updating! Bit ocurripied last days. Will take a look tmr.

src/blob/handlers/BlobHandler.ts
XiaoningLiu: yes. `AuthenticationFailed` and `AuthorizationFailure` all returns status code 403. there are some differents for error messages between azurite and azure storage.

PR #547, opened 2020-09-03
New, leaner Dockerfile

ItalyPaleAle:
This contains a new Dockerfile which allows creating much smaller Docker images and with less layers.

It uses multi-stage builds, so tsc runs in a different container (bonus: it now runs "npm run test" too).

Also, updated .dockerignore that was preventing the container from copying the self-signed cert used by tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,42 @@
+#
+# Builder
+#
+FROM node:lts-alpine AS builder
+
+WORKDIR /opt/azurite
+
+# Install dependencies first
+COPY *.json LICENSE NOTICE.txt ./
+RUN npm config set unsafe-perm=true && \
+  npm ci
+
+# Copy the source code and build the app
+COPY src ./src
+COPY tests ./tests
+RUN npm run build && \
+  npm run test && \
+  npm install -g --loglevel verbose
+
+
+#
+# Production image
+#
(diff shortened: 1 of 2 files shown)


ItalyPaleAle:
For some data on the improvements:

**Current Docker image**

- Size: 303 MB
- Layers: 14

**New Docker image**

- Size: 129 MB (-58%)
- Layers: 9 (-36%)

Dockerfile
```diff
@@ -1,20 +1,43 @@
...
-COPY tests ./tests
-COPY *.json ./
-COPY LICENSE .
-COPY NOTICE.txt .
+COPY package*.json ./
+
+RUN npm config set unsafe-perm=true && \
+  npm ci
+
+COPY --from=builder /opt/azurite/dist/ dist/
```
XiaoningLiu, line 39: We need LICENSE and NOTICE.txt file part of final image too.

XiaoningLiu (commented):
That a great improvement! Generally looks good. We need to make sure license and notice files part of final image too.

ItalyPaleAle:
@XiaoningLiu thanks for the review! I've made the change as requested.

XiaoningLiu:
@ItalyPaleAle Thanks!

@blueww Can you help review?

PR #549, opened 2020-09-03
Allow canonicalized prefix for blob metadata operations

jwwk:
Recently, I have run into a problem using the official Golang SDK for Azure Blob Storage and the Azurite emulator.
I have previously discussed the issue over at the SDK's repo (see https://github.com/Azure/azure-storage-blob-go/issues/191).

The production service handles metadata operations slightly differently than the service emulator, as it accepts set-metadata operations using a canonicalized header prefix (`X-Ms-Meta-` instead of the documented `x-ms-meta-`).

I guess it's best to stick closer to the production service than to the literal documentation in this case. I've updated the respective part of Azurite's metadata header algorithm to accept the same headers that the production service accepts.

### DIFF
--- a/src/common/utils/utils.ts
+++ b/src/common/utils/utils.ts
@@ -28,7 +28,7 @@ export function convertRawHeadersToMetadata(
   for (let i = 0; i < rawHeaders.length; i = i + 2) {
     const header = rawHeaders[i];
     if (
-      header.startsWith(metadataPrefix) &&
+      header.toLowerCase().startsWith(metadataPrefix) &&
       header.length > metadataPrefix.length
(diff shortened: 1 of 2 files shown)


blueww:
@jwwk 
Please make the CI pass.
Is the change to align with server behavior?

@XiaoningLiu 
Are you OK with the change?

jwwk:
@blueww I assume one of the CI tasks timed out. Maybe someone can just restart the process.
As mentioned in the PR, this change is to make Azurite behave more like the real Azure.

XiaoningLiu (approved):

PR #550, opened 2020-09-05
490 merge entity

iamNoah1:


src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -199,12 +199,37 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
...
+    if (!currentDoc) {
+      throw StorageErrorFactory.getEntityNotExist(context);
+    } else {
+      // Test if etag value is valid
+      if (etag !== "*" && currentDoc.eTag !== etag) {
+        throw StorageErrorFactory.getPreconditionFailed(context);
+      }
+    }
+
+    tableColl.update(currentDoc);
```
XiaoningLiu, line 232: >tableColl.update(currentDoc); [](start = 4, length = 29)

Seems we don't have 'merge' happens here. 

The merge should merge current doc properties with incoming entities. For example,

const mergedDoc = {
  ...currentDoc,
  ...entity,
  properties: {
    ..currentDoc.properties,
    ..entity.properties
  }
}

tests/table/apis/table.entity.test.ts
```diff
@@ -316,4 +316,147 @@ describe("table Entity APIs test", () => {
       }
     );
   });
+
+  it("Merge an Entity that exists, @loki", done => {
```
XiaoningLiu, line 320: >(" [](start = 4, length = 2)

Table doesn't have "queryEntities" added, so I know we currently cannot verify a merged entity. Can you add some TODO after mergeentity call to validate the entity merge is actually successful?

XiaoningLiu:
Hi @iamNoah1, still are some unresolved comments. Edwin and Martha is now working on Insert or Merge entity, after that there maybe conflicts to merge current PR.

iamNoah1:
@XiaoningLiu will investigate on the comments the next days.

src/table/persistence/LokiTableMetadataStore.ts
iamNoah1: I don't know how I could oversee that. Added it. Manually checked that all properties of entity override properties of currentDoc. Modification time gets updated, new etag is generated. Or did you mean by double checking really fetching the doc again in between the time of the first fetch and the merge operation of tableColl to be sure that is was not modified in the meantime?

tests/table/apis/table.entity.test.ts
iamNoah1: Done.
Im just thinking out loud here, but as we use loki in memory db, would it make sense to check directly in the database if the entity was merged properly? Otherwise we would rely on the correctness of queryEntities function to verify that the entity was merged correctly.

src/table/persistence/LokiTableMetadataStore.ts
XiaoningLiu: Ok. Just to check whether we need to generate a new etag and timestamp for the merged entity.

By the way, If we do a simple 
```
const mergedDoc = {
      ...currentDoc,
      ...entity
    };
```
Then old entity's properties will be override by the new entity.properties. If we want the merge happen for properties, we need to do something like:
```javascript
 const mergedDoc = {
  ...currentDoc,
  ...entity,
  properties: {
    ...currentDoc.properties,
    ...entity.properties
  }
}
```

---
In reply to: [493550304](https://github.com/Azure/Azurite/pull/550#discussion_r493550304) [](ancestors = 493550304)

tests/table/apis/table.entity.test.ts
XiaoningLiu: Possible, The thing is how we can access loki instance directly from a test case.

iamNoah1: Good question. Adding a method to ITableMetaDataStore which returns the real db reference and exposing it in the TableServer. But I am not sure if that is a feasible way to go or worth it at all?!

src/table/persistence/LokiTableMetadataStore.ts
iamNoah1: updated it

iamNoah1:
@XiaoningLiu What should we do with Merge Conflicts?

tests/table/apis/table.entity.test.ts
XiaoningLiu: TableServer 

> @XiaoningLiu What should we do with Merge Conflicts?

Afarid we need to do some manual merge work to resolve the conflicts. There are some comment logic between this PR and Insert Or Merge, so needs to filter out the comment parts and non-comment parts. Sorry for the inconvience.

XiaoningLiu:
> @XiaoningLiu What should we do with Merge Conflicts?

Afarid we need to do some manual merge work to resolve the conflicts. There are some comment logic between this PR and Insert Or Merge, so needs to filter out the comment parts and non-comment parts. Sorry for the inconvience.

iamNoah1:
@XiaoningLiu thanks for the patience. I finally solved mere conflicts. Hope that everything is fine now. Let me know if there is anything that still has to be adjusted

iamNoah1:
@XiaoningLiu let me know we should go on from here

XiaoningLiu:
Seems Merge Entity is already there.

PR #551, opened 2020-09-07
Add queryEntities

isabella0428:
Add queryEntities, also modify insertEntity (store odata with inserted entity)


XiaoningLiu:
Hi @isabella0428, can you add test cases?

src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -124,11 +125,109 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
...
+        }
+
+        // Add to result
+        selectedFieldsResult.push(entitySelectedFieldResult);
+      }
+
+      selectedResult = selectedFieldsResult;
+    }
+
+    if (queryOptions.top !== undefined) {
```
jikuja, line 226: The default value for $top should be 1000 for feature parity with Azure Table service.

ref: https://docs.microsoft.com/en-us/rest/api/storageservices/query-entities#remarks

src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -124,11 +125,109 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
...
+      selectedResult = selectedResult.slice(0, queryOptions.top!);
```
jikuja, line 227: I think for best performance it is best to use Loki's ResultSet chaining instead of creating a new full array of data just to splice then to user requested amount. Something like:

`const result = tableColl.chain().find(filterJson).limit(valueOfTopOr100).data()`

XiaoningLiu:
Merge then resolve comments.

PR #552, opened 2020-09-08
[Table] Phase 1 Feature - Insert Or Replace Entity

edwin-huber:
[Table] Phase 1 Feature - Insert Or Replace Entity #492

Upsert added incl query table entities with partition key and row key in the loki data handler.

Added test case to validate "insert" API call via SDK, and verified that this does not include the ifMatch / etag header on the request to the API.

### DIFF
--- a/src/table/handlers/TableHandler.ts
+++ b/src/table/handlers/TableHandler.ts
@@ -14,7 +14,7 @@ import {
   NO_METADATA_ACCEPT,
   RETURN_CONTENT,
   RETURN_NO_CONTENT,
-  TABLE_API_VERSION,
+  TABLE_API_VERSION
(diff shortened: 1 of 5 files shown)


edwin-huber:
Would use query entities / query with partition key and row key to extend tests to validate the entities placed in table storage.

XiaoningLiu (commented):
Hi @Edwin, the changes looks good to me. Can you check the CI failures?

XiaoningLiu:
> Would use query entities / query with partition key and row key to extend tests to validate the entities placed in table storage.

We can leave a TODO in test cases, or add an empty test cases for now. Query entity is under review.

XiaoningLiu (approved):

PR #555, opened 2020-09-12
Add API Version check for table storage middleware

### LINKED ISSUE #516: [Table] Phase 1 Feature - API Version Check
Same with Blob/Queue. Validate incoming RESTFul API version is within allowed list.
Support skipAPIVersionCheck parameter

ocdi:
Fixes #516. The implementation is 90% copy/paste from the equivalent methods in blob storage.

### DIFF
--- a/src/table/TableRequestListenerFactory.ts
+++ b/src/table/TableRequestListenerFactory.ts
@@ -32,7 +32,8 @@ export default class TableRequestListenerFactory
   public constructor(
     private readonly metadataStore: ITableMetadataStore,
     private readonly enableAccessLog: boolean,
-    private readonly accessLogWriteStream?: NodeJS.WritableStream // private readonly skipApiVersionCheck?: boolean,
+    private readonly accessLogWriteStream?: NodeJS.WritableStream,
+    private readonly skipApiVersionCheck?: boolean
   ) {}
 
   public createRequestListener(): RequestListener {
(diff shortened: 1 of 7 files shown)


XiaoningLiu (changes requested):
Thanks @ocdi! 

As you noticed, `skipApiVersionCheck` is a flag parameter to indicate whether we should enable the api version check. The parameter is parsed from `azurite` or `azurite-table` commandline parameter, or from Visual Studio Code settings. Finally the parameter is passed into `tableStorageContextMiddleware` you modified already.

Can you do a manual check to verify `skipApiVersionCheck` commandline parameter works for `azurite` and `azurite-table` commandline? (Table doesn't support Visual Studio Code yet, we can ignore Visual Studio Code)

Seems the `skipApiVersionCheck` commandline parameter is supported, but doesn't pass into the `tableStorageContextMiddleware` yet.

Tips: We can globally search keyword "skipApiVersionCheck" and compare any missing parts between blob/queue with table.

ocdi:
@XiaoningLiu thanks for the tips here. I added the switch to `launch.json` and passed it through the same way as blob did it. I had missed those last two bits, and was unsure how to properly test that at the time. 

I modified one of the Postman requests and set the header to an invalid value and debugged to verify.

image

XiaoningLiu:
Thanks! Looks good.

Issue #443, opened 2020-04-28
Fix blob copy support when using SQL metadata implementations


atrauzzi:
### Which service(blob, file, queue, table) does this issue concern?
blob
### Which version of the Azurite was used?
`latest`
### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
DockerHub
### What's the Node.js version?
N/A
### What problem was encountered?
Can't perform simple copy operations.
### Steps to reproduce the issue?
Attempt to call copy using Azure Storage Explorer or your favourite SDK, you'll get a 500.
### Have you found a mitigation/solution?
No.

atrauzzi:
Just as a note, I think this functionality is too essential to not be supported.  This was a landmine for me today.

blueww:
@atrauzzi 
We do have Test case to cover the blob copy scenarios.
So pleases give the detail repro steps, like how to prepare the blob to copy, the parameter you use in copy, the detail error message, and better with Azurite debug log.  
These information can help us to better investigate this issue.

atrauzzi:
Hey @blueww!

I'm not sure it's necessary as [you guys have already acknowledged that this is an issue](https://github.com/Azure/Azurite/commit/a75984b6ff2dc569b2c8c07d0446720a8067484d#diff-04c6e90faac2675aa89e2176d2eec7d8R363) in the readme.

![image](https://user-images.githubusercontent.com/115199/80663519-d0eae980-8a59-11ea-9b91-209311325ae6.png)

The repro steps are to enable SQL based metadata and to perform a copy using Azure Storage Explorer or an SDK.  I have used no special parameters.  It's just the most basic copy operation you can think of.

blueww:
@atrauzzi 
Yes, we currently don't support copy page/append blob on SQL based metadata implementation.
This is in our track. we will prioritize it with other feature ask.

atrauzzi:
Much appreciated! :smile:

patrickcarnahan:
@atrauzzi which sql dialect are you using?

atrauzzi:
Postgres.

patrickcarnahan:
@atrauzzi #577 adds support for asynchronous copy using the `sql` metadata store. it's entirely possible that the UI makes use of the synchronous copy call, i'm not positive how it works. 

care to give it a spin and see if it resolves you issue?

atrauzzi:
I'll definitely take a look next chance I get and report back!  Thank you! :smile:

PR #577, opened 2020-09-19
enable async copy for sql metadata store

patrickcarnahan:
this adds missing functionality to the sql metadata store to enable asynchronous blob copy operations. it has full parity with the `loki` store, so the tests have been enabled for both `@loki @sql` where appropriate.

fixes https://github.com/Azure/Azurite/issues/443.

### DIFF
--- a/src/blob/persistence/SqlBlobMetadataStore.ts
+++ b/src/blob/persistence/SqlBlobMetadataStore.ts
@@ -2467,7 +2467,7 @@ export default class SqlBlobMetadataStore implements IBlobMetadataStore {
     }
 
     const blobType = this.getModelValue<Models.BlobType>(res, "blobType", true);
-    const isCommitted = this.getModelValue<boolean>(res, "iscommitted", true);
+    const isCommitted = this.getModelValue<boolean>(res, "isCommitted", true);
 
     return { blobType, isCommitted };
   }
--- a/src/blob/persistence/SqlBlobMetadataStore.ts
+++ b/src/blob/persistence/SqlBlobMetadataStore.ts
@@ -11,6 +11,8 @@ import {
   Transaction
 } from "sequelize";
 
+import uuid from "uuid/v4";
+
 import {
   DEFAULT_SQL_CHARSET,
   DEFAULT_SQL_COLLATE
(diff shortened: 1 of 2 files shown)


src/blob/persistence/SqlBlobMetadataStore.ts
```diff
@@ -3209,4 +3201,106 @@ export default class SqlBlobMetadataStore implements IBlobMetadataStore {
     }
     return { lease: leaseString };
   }
+
+  private async getBlobWithLeaseUpdated(
```
patrickcarnahan, line 3205: this looked like a useful method that only existed in the `loki` store, so for now it has been ported to use `sequelize ` for CRUD. there is a much larger potential for refactoring here i think, but i will take that up separately.

src/blob/persistence/SqlBlobMetadataStore.ts
```diff
@@ -1468,47 +1428,41 @@ export default class SqlBlobMetadataStore implements IBlobMetadataStore {
     });
   }
 
-  public async getBlockList(
+  public getBlockList(
```
patrickcarnahan, line 1431: this method actually had a bug. it was returning an empty block list instead of throwing `BlobNotFound` when the requested blob didn't exist. i will be adding a test for this case along with filing an issue.

src/blob/persistence/SqlBlobMetadataStore.ts
```diff
@@ -1122,17 +1113,12 @@ export default class SqlBlobMetadataStore implements IBlobMetadataStore {
     modifiedAccessConditions?: Models.ModifiedAccessConditions
   ): Promise<void> {
     return this.sequelize.transaction(async t => {
-      const containerFindResult = await ContainersModel.findOne({
```
patrickcarnahan, line 1125: this code appeared over and over so with the addition of the new method `assertContainerExists` i removed this specific pattern throughout the file.

src/blob/persistence/SqlBlobMetadataStore.ts
```diff
@@ -2632,7 +2459,7 @@ export default class SqlBlobMetadataStore implements IBlobMetadataStore {
     }
 
     const blobType = this.getModelValue<Models.BlobType>(res, "blobType", true);
-    const isCommitted = this.getModelValue<boolean>(res, "iscommitted", true);
+    const isCommitted = this.getModelValue<boolean>(res, "isCommitted", true);
```
patrickcarnahan, line 2462: this caused the API to return a `500` since the property wasn't found on the object.

src/blob/persistence/SqlBlobMetadataStore.ts
patrickcarnahan: this fix was actually moved to #578.

PR #578, opened 2020-09-21
fix sql metadata store getBlockList for non-existent blob

patrickcarnahan:
currently there is an issue with the `sql` metadata store and `getBlockList`. this API behaves differently between `loki` and `sql`, and the `loki` behavior appears to be correct when compared to real azure storage.

a test has been added to illustrate the issue, and a secondary commit will be added to validate the fix.

fixes https://github.com/Azure/Azurite/issues/579


patrickcarnahan:
@XiaoningLiu this is separated from the other PR but I can also combine if that's preferable. Currently there is an ordering dependency between changes unless this is taken as a fix to #577.

XiaoningLiu (approved):

PR #580, opened 2020-09-27
Handle SIGTERM. Signal for Docker stop

andymac4182:
Docker defaults to using SIGTERM on `docker stop` This consolidates the handlers to a single function used by the 3 ways of stopping the app.

You can see this by the time it takes to stop the Azurite container without handling `SIGTERM` vs when you handle it. Docker will wait 10 seconds before forcibly killing the container. 

### Latest container
```
❯ time docker stop azurite
azurite

real    0m10.467s
user    0m0.073s
sys     0m0.100s
```

### Container that handles SIGTERM
```
❯ time docker stop azurite-sigterm
azurite-sigterm

real    0m0.431s
user    0m0.111s
sys     0m0.053s
```

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,6 +5,7 @@
 ## Upcoming Release
 
 - Add support for copying blobs across storage accounts within the same Azurite instance.
+- Add handling of SIGTERM to gracefully stop the docker container
 
 ## 2020.07 Version 3.8.0
 
--- a/src/azurite.ts
+++ b/src/azurite.ts
@@ -16,10 +16,32 @@ import {
   DEFAULT_QUEUE_PERSISTENCE_ARRAY,
   DEFAULT_QUEUE_PERSISTENCE_PATH
 } from "./queue/utils/constants";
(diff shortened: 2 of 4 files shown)


blueww:
@andymac4182 
Thanks for raising the PR!
It generally look good to me.

Would you please also update the change log https://github.com/Azure/Azurite/blob/master/ChangeLog.md, then we can merge the PR.

blueww (changes requested):
Please update the change to https://github.com/Azure/Azurite/blob/master/ChangeLog.md#upcoming-release

andymac4182:
Thanks @blueww. I have updated the changelog now

blueww:
@andymac4182 
Thanks for the update!
Will merge after CI pass.

andymac4182:
Thanks. It looks like CI has gotten stuck. It might need that docker step reran.

andymac4182:
Yeah. Fixed now it has re ran :+1:

blueww (approved):

PR #624, opened 2020-11-15
Use un-url-encoded blobName value, when calculate SAS on the check

### LINKED ISSUE #415: Blob SAS fails if query string contains URL-encoded value

blob (but may be applied to queue too)

v3.5.0

DockerHub

v10.19.0

SAS authentification for blob fails if SAS contains URL-encoded strings (e.g. URL-encoded filename in Content-Disposition)

I used python client for a minimal reproducible code,

```python
import datetime
import urllib.parse

import requests
from azure.storage.blob.blockblobservice import BlockBlobService
from azure.storage.blob.models import BlobPermissions, ContentSettings

azure_blob_service = BlockBlobService(...)

name = 'te st.txt'
name_quoted = urllib.parse.quote(name)  # will result 'te%20st.txt'

# assume that the file is already uploaded in the blob storage

sas_token = azure_blob_service.generate_blob_shared_access_signature(
    'test',
    name,
    permission=BlobPermissions(read=True),
    content_disposition=f'inline; filename="{name_quoted}"',
    expiry=datetime.datetime(2030, 1, 1)
)

url = azure_blob_service.make_blob_url(
    'test',
    name_quoted,
    sas_token=sas_token
)

print(requests.get(url))
# In real Azure Blob Storage, this results 200 OK response,
# In Azurite, this results 403 Forbidden response.
```

I guess that in `BlobSASAuthenticator.ts`, query values (which are fetched from `req.getQuery(...)`) are already decoded by `express` and `qs` packages but `this.decodeIfExist(...)` function applies `decodeURIComponent` one more time, which results in different SAS signature values.

For example, `contentDisposition` value from `rscd` query key,
- expected : `inline; filename="te%20st.txt"`
- actual : `inline; filename="te st.txt"`

iBavtovich:
Fixes #415

When Azure SDK works with blob and its name contains symbols, which should be encoded in `url`, signature calculation performed from the un-url-encoded `blobName` value. In order to SAS check in Azure passes, we need to calculate SAS under the hood in the same way.

### DIFF
--- a/src/blob/authentication/BlobSASAuthenticator.ts
+++ b/src/blob/authentication/BlobSASAuthenticator.ts
@@ -57,7 +57,7 @@ export default class BlobSASAuthenticator implements IAuthenticator {
       return undefined;
     }
 
-    const blobName = blobContext.blob;
+    const blobName = this.decodeIfExist(blobContext.blob);
     this.logger.debug(
       // tslint:disable-next-line:max-line-length
       `BlobSASAuthenticator:validate() Retrieved account name from context: ${account}, container: ${containerName}, blob: ${blobName}`,


XiaoningLiu:
Thanks @iBavtovich! A test case failed the pipeline, can you help check?

https://dev.azure.com/azure/Azurite/_build/results?buildId=16293&view=logs&jobId=5c249ced-33d3-5644-8846-374752928f2b&j=5c249ced-33d3-5644-8846-374752928f2b&t=da461190-4564-5ab0-1aee-1dde69759da2

iBavtovich:
I double checked the problem and found that the request url was url-encoded twice: 1 time by Azure SDK and second by the http client library I used (which is wrong). But on the Azurite side it was decoded once (as should be)  in `blobStorageContextMiddleware` middleware. So there is no need to decode `blobName` again in `BlobSASAuthenticator`.

PR #666, opened 2021-01-04
Fault injection

ljian3377:


XiaoningLiu:
>     res.getBodyStream().write(xmlBody);

As we synced offline. We can consider introduce a callback, for example, callback in context like context.serializeInject to hijack the response stream.

---
Refers to: src/blob/generated/utils/serializer.ts:316 in 1c05fb2. [](commit_id = 1c05fb297432b1efcf119d33ed2bed0013f75663, deletion_comment = False)

PR #674, opened 2021-01-13
Discussion PR - Integration tests for entity group transactions

edwin-huber:
As discussed, we have separated out the initial integration tests for entity group transactions / batch into a separate PR to allow some parallel work.

Suggestion is that the tests are written against the service, and then refactored to Azurite.

Serialization and De-Serialization of batch requests are covered by unit tests and will be submitted with the PR for table batch.

### DIFF
--- a/tests/table/apis/table.entity.test.ts
+++ b/tests/table/apis/table.entity.test.ts
@@ -16,6 +16,17 @@ import {
 // Set true to enable debug log
 configLogger(false);
 
+// Create Entity for tests
+function createBasicEntityForTest() {
+  return {
+    PartitionKey: eg.String("part1"),
+    RowKey: eg.String(getUniqueName("row")),
+    myValue: eg.String("value1")
+  };
+}
+
+const eg = Azure.TableUtilities.entityGenerator;
+
 describe("table Entity APIs test", () => {
   // TODO: Create a server factory as tests utils
   const protocol = "http";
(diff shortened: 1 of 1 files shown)


tests/table/apis/table.entity.test.ts
```diff
@@ -436,4 +447,130 @@ describe("table Entity APIs test", () => {
       }
     );
   });
+
+  it.only("Simple batch test: Inserts multiple entities as a batch, @loki", (done) => {
+    requestOverride.headers = {
+      Prefer: "return-content",
+      accept: "application/json;odata=fullmetadata"
+    };
+    const batchEntity1 = createBasicEntityForTest();
```
XiaoningLiu, line 456: Table requires unique combinations of "PartitionKey" and "RowKey" for different entities.

XiaoningLiu (approved):
One comment added.

We can merge this to unblock others keeping contributing to Batch test cases.

Seems pipeline failed with the change, can we make pipeline green?

edwin-huber:
Hi,  
I submitted this as a PR for discussion, as the point of the integration tests is that they will fail until we get the implementation right...  
I see I forgot the "only" designation on the insert test, but all tests for batch will fail.  
I'm happy to collaborate with whomever is writing the tests, but we would need to invalidate the tests if we want the builds to complete, which removes the point of committing the tests until batch is ready.

Your thoughts?

XiaoningLiu:
> writing the tests, but we would need to invalidate the tests if we want the builds to complete, which removes the point of committing the tests until batch is ready.

Ok to leave the PR here with cases failures. @tzhanl can you help pick up from here?

tests/table/apis/table.entity.test.ts
edwin-huber: Thanks, yes.  
This is why the helper function creates a new row for each entity 😸 
```javascript
getUniqueName("row")
```

edwin-huber:
Now implemented and merged in table branch.

Issue #303, opened 2019-12-16
StageBlock/UploadPages: not check the content aligned with "Content-MD5" header


blueww:
### Which service(blob, file, queue, table) does this issue concern?
blob
### Which version of the Azurite was used?
3.3.0-preview
### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
VS code
### What's the Node.js version?
v10.15.3
### What problem was encountered?
StageBlock/UploadPages: not check the content aligned with "Content-MD5" header
So the request will success, even input a wrong  "Content-MD5" header
### Steps to reproduce the issue?
StageBlock/UploadPages with a wrong a wrong  "Content-MD5" header
### Have you found a mitigation/solution?
No

Rest doc:

> Content-MD5 | Optional. An MD5 hash of the page content. This hash is used to verify the integrity of the page during transport. When this header is specified, the storage service compares the hash of the content that has arrived with this header value.

Test case info:

> Test Name:	ExtendedErrorInfoVerifyXmlWithAdditionalDetails
> Test FullName:	Microsoft.Azure.Storage.StorageExceptionTests.ExtendedErrorInfoVerifyXmlWithAdditionalDetails
> Test Source:	C:\code\azure-storage-net\Test\ClassLibraryCommon\StorageExceptionTests.cs : line 137
> Test Outcome:	Failed
> Test Duration:	0:00:01.6398374
> 
> Result StackTrace:	
> at Microsoft.Azure.Storage.TestHelper.ExpectedException[T](Action operation, String operationDescription, String expectedMessage) in C:\code\azure-storage-net\Test\Common\TestHelper.Common.cs:line 85
>    at Microsoft.Azure.Storage.StorageExceptionTests.ExtendedErrorInfoVerifyXmlWithAdditionalDetails() in C:\code\azure-storage-net\Test\ClassLibraryCommon\StorageExceptionTests.cs:line 164
> Result Message:	Assert.Fail failed. No exception received while expecting Microsoft.Azure.Storage.StorageException: Invalid MD5 should fail with mismatch

torbjornvatn:
I also experienced this @blueww when my tests relying on Azurite started failing.
I use the https://godoc.org/github.com/Azure/azure-storage-blob-go/azblob#UploadStreamToBlockBlob function from https://github.com/Azure/azure-storage-blob-go

blueww:
The check is still not implemented in Azurite. 
We will priority it with all other feature requirements.

[...]

PR #689, opened 2021-01-28
BlockBlob_StageBlock now checks Content-MD5 and errors if there is an MD5-mismatch

wayne-li2:
BlockBlob_StageBlock now checks Content-MD5 and errors if there is an MD5-mismatch.

This PR tackles issue #303 only for BlockBlob_StageBlock.


blueww (approved):

blueww:
@XiaoningLiu 
The PR is OK for me.
Do you have any comments?

XiaoningLiu:
It looks good to me and it should work. There is still some space we can do to improve the performance. The implementation now adds additional reading from disk to calculate the md5. Ideally we can do the md5 calculation during the first extentAppend operation on the air. We can reduce one additional disk read. This is an advanced ask, just to check if @wayne-li2 you want to check this? : )

wayne-li2:
@XiaoningLiu I could only spare a brief look today, and am afraid I can't commit to more time right now. Would it be possible to merge this + release a new Docker image? Happy to continue being a contributor in the future!

wayne-li2:
@blueww @XiaoningLiu Bump on this PR? Also, there seems to be a failing check, but looking at the output it is unrelated to any changes in this PR.

PR #691, opened 2021-01-29
Table Batch / Entity Group Transactions v1

edwin-huber:
Unit Tests and Integration Tests added.

ToDos:
```
src\table\batch\BatchSerialization.ts
```
 - Validate behvaiour for GET HTTP verb / Query op
 - Improve RegEx
```
src\table\batch\TableBatchManager.ts
```
 - Implement all APIs used by batch / entity group transactions
```
General
```
 - Validate behaviour matches service in failure cases.
 - Validate header behaviors and requirements
 - Include tests from @zzhxiaofeng 
 - Correct handling of Content ID as per OData specification.

### Development Guideline


edwin-huber:
Seems we have issues with node 8x and the new code... need to check this by installing an old version of node.

edwin-huber:
There is still quite a bit of refactoring coming to improve the serialization code.  
This is dependent on the completion of test coverage and query handling.  
We are still working on :
- ensuring unit tests for serialization are correctly checking all response types.
- covering all query types

package.json
```diff
@@ -67,6 +68,7 @@
     "mocha": "^5.2.0",
     "prettier": "^2.1.2",
     "prettier-tslint": "^0.4.2",
+    "to-readable-stream": "^2.1.0",
```
XiaoningLiu, line 71: Only need to add this dependency into "dependencies". Don't need in "devDependencies".

src/table/handlers/TableHandler.ts
```diff
@@ -245,6 +249,8 @@ export default class TableHandler extends BaseHandler implements ITableHandler {
   }
 
   // TODO: Create data structures to hold entity properties and support serialize, merge, deserialize, filter
+  // Note: Batch is using the partition key and row key args, handler receives these values from middleware via
+  // context
```
XiaoningLiu, line 253: minor: Change _partitionKey to partitionKey if we actually use it. Same with other variables.

src/table/batch/TableBatchOrchestrator.ts
```diff
@@ -0,0 +1,533 @@
...
+   * see Link below for details of response format
+   * tslint:disable-next-line: max-line-length
+   * https://docs.microsoft.com/en-us/rest/api/storageservices/performing-entity-group-transactions#json-versions-2013-08-15-and-later-2
+   *
+   * @private
+   * @return {*}  {string}
+   * @memberof TableBatchManager
+   */
+  private serializeResponses(): string {
+    let responseString: string = "";
```
XiaoningLiu, line 104: If batch suboperation includes query, then one and only one query suboperation is allowed in batch.
We can check if Logic App/Durable Function demo requires this. Should be not.

src/table/handlers/TableHandler.ts
edwin-huber: done, removed from all handler functions

src/table/batch/TableBatchOrchestrator.ts
edwin-huber: Adding as a work item

edwin-huber:
Discussed and approved with @XiaoningLiu

Issue #688, opened 2021-01-28
TableStorage returns enparseable etag for Insert operation


mstrejczek:
### Which service(blob, file, queue, table) does this issue concern?
Table

### Which version of the Azurite was used?
3.9.0-table-alpha.1

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
NPM

### What's the Node.js version?
v14.15.4

### What problem was encountered?
Insert entity operation fails when using Azure Storage java library.
IllegalArgumentException is thrown from `TableResult.parseETagForTimestamp(String etag)`
Debugger indicates that etag value returned by Azurite was "0x205B6F86DEB6480".

### Steps to reproduce the issue?
Fails every time for me whenever I add an entity to a table.
I use `com.microsoft.azure:azure-storage:8.6.5` library - works fine with Azure Table Storage in the cloud.

### Have you found a mitigation/solution?
Nope

blueww:
@mstrejczek 
Currently the Etag generated from Azurite is a string that only matches etag format, it doesn't contain timestamp.
see code https://github.com/Azure/Azurite/blob/aa6cba7904e0c6e147bf1f932981776a09d7d059/src/common/utils/utils.ts#L50

We welcome contribution.
It would be great if you could help to fix the generate etag function!

mstrejczek:
I've just checked how etag returned by Azure Table Storage looks like and it is something like
`W/"datetime'2021-02-01T10%3A58%3A01.2219993Z'"`
Basically a constant 
`W/"datetime'`
followed by ISO8601 datetime and then followed by another constant:
`"'`
I'll try to put together a PR when I have some time.

mstrejczek:
This ticket can be reproduced with this test app: https://github.com/mstrejczek/azurite-table-test

blueww:
@mstrejczek 
Thanks for the reply!
It would be great if you can raise a PR to fix it.
Please note, as the currently newEtag() if also used by Blob storage in Azurite, would you please write a new function to create etag for table request, and update table APIs to use your new function.

PR #697, opened 2021-02-05
Etag returned by Table Storage follows Azure format now

mstrejczek:
Fix for https://github.com/Azure/Azurite/issues/688 (format of etag returned by Table Storage endpoint)

### DIFF
--- a/src/common/utils/utils.ts
+++ b/src/common/utils/utils.ts
@@ -60,9 +60,9 @@ export function newEtag(): string {
   );
 }
 
-export function newTableEntityEtag(): string {
+export function newTableEntityEtag(startTime: Date): string {
   // Etag as returned by Table Storage should match W/"datetime'<ISO8601datetime>'"
-  return "W/\"datetime'" + new Date().toISOString() + "'\"";
+  return "W/\"datetime'" + truncatedISO8061Date(startTime, true) + "'\"";
 }
 
 /**
--- a/src/common/utils/utils.ts
+++ b/src/common/utils/utils.ts
@@ -60,6 +60,11 @@ export function newEtag(): string {
   );
 }
 
+export function newTableEntityEtag(): string {
+  // Etag as returned by Table Storage should match W/"datetime'<ISO8601datetime>'"
+  return "W/\"datetime'" + new Date().toISOString() + "'\"";
+}
+
 /**
  * Generates a hash signature for an HTTP request or for a SAS.
  *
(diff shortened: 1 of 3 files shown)


src/common/utils/utils.ts
```diff
@@ -60,6 +60,11 @@ export function newEtag(): string {
   );
 }
 
+export function newTableEntityEtag(): string {
+  // Etag as returned by Table Storage should match W/"datetime'<ISO8601datetime>'"
+  return "W/\"datetime'" + new Date().toISOString() + "'\"";
```
XiaoningLiu, line 65: Thanks for contribution! Entity etag uses datetime aligns with entity lastmodifiedtime. Can you help update this method, accept a datetime parameter. Like this:

```javascript
    const eTag = newTableEntityEtag(context.startTime!);

    const entity: Entity = {
      PartitionKey: partitionKey,
      RowKey: rowKey,
      properties: options.tableEntityProperties,
      lastModifiedTime: context.startTime!,
      eTag
    };
```

XiaoningLiu: Also please increase the datetime precision like "2021-01-27T09:08:45.3301416Z". By default, precision of toISOString() is not good enough.

```
new Date().toISOString()
'2021-02-09T07:57:02.680Z'
```

mstrejczek: Updated, thanks for comments!
7 millisecond digits precision is a stretch since Date() object has precision of three ms digits. So precision now is high but accuracy same as before.

XiaoningLiu:
@mstrejczek Forgot about the etag encoding before click the merge button : ) Can you help make an update to url encode the etag in `newTableEntityEtag` method?

Typically, etag in following response body 'odata.etag' and 'ETag' header should be align with encoded pattern like `W/"datetime'2021-02-10T02%3A49%3A22.8458722Z'"`.
- Query Entity
- Query Entity with ParitionKey & RowKey
- Insert Entity
- Updat Entity

tests/table/apis/table.entity.test.ts
```diff
@@ -75,6 +75,11 @@ describe("table Entity APIs test", () => {
     };
     tableService.insertEntity(tableName, entity, (error, result, response) => {
       assert.equal(response.statusCode, 201);
+      assert.ok(
+        response.headers?.etag.match(
+          "W/\"datetime'\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}.\\d{7}Z'\""
```
XiaoningLiu, line 80: We should expect something like `W/"datetime'2021-01-27T09%3A08%3A45.3301416Z'"`. Pay attention to `:` is encoded to `%3A`

mstrejczek: Good point about encoding, I'll submit another PR to fix this.

PR #698, opened 2021-02-05
Skip Content-Length check for Append Block if --loose is set

### LINKED ISSUE #682: AppendBlock from azure-sdk-for-go v50.1.0 doesn't work

Blob

Latest (v3.10.0)

DockerHub (`mcr.microsoft.com/azure-storage/azurite:3.10.0`)

Whatever the `mcr.microsoft.com/azure-storage/azurite:3.10.0` docker container is running

[`Blob. AppendBlock`](https://github.com/Azure/azure-sdk-for-go/blob/ac17097a799300c6446ec0e3f899a4f0b2a81375/storage/appendblob.go#L70) from `github.com/Azure/azure-sdk-for-go` `v3.10.0` fails with error: `Content-Length must be 0 for Create Append Blob request.`

Run the following code:

```go
package main

import (
	"log"
	"time"

	"github.com/Azure/azure-sdk-for-go/storage"
	"github.com/ory/dockertest/v3"
	"github.com/ory/dockertest/v3/docker"
)

// RunAzurite starts an azurite container
func RunAzurite(pool *dockertest.Pool) (*dockertest.Resource, error) {
	opts := dockertest.RunOptions{
		Repository:   "mcr.microsoft.com/azure-storage/azurite",
		Tag:          "3.10.0",
		ExposedPorts: []string{"10000"},
		PortBindings: map[docker.Port][]docker.PortBinding{
			"10000": {{HostIP: "0.0.0.0", HostPort: "10000"}},
		},
	}
	azurite, err := pool.RunWithOptions(&opts)
	if err != nil {
		return nil, err
	}
	if eerr := azurite.Expire(10); eerr != nil {
		return nil, eerr
	}
	pool.MaxWait = 10 * time.Second
	rerr := pool.Retry(func() error {
		client, eerr := storage.NewEmulatorClient()
		if eerr != nil {
			return eerr
		}
		s := client.GetBlobService()
		c := s.GetContainerReference("cont")
		if _, err = c.Exists(); err != nil {
			return err
		}
		return nil
	})
	return azurite, rerr
}

func main() {
	pool, err := dockertest.NewPool("")
	if err != nil {
		log.Fatalf("Failed to create dockertest pool: %s", err)
	}

	azurite, err := RunAzurite(pool)
	defer pool.Purge(azurite)

	if err != nil {
		log.Fatalf("Failed to start Azurite: %s", err)
	}

	client, err := storage.NewEmulatorClient()
	if err != nil {
		log.Fatalf("Failed to create storage client: %s", err)
	}

	blobClient := client.GetBlobService()
	dummyContainer := "foobar"
	containerRef := blobClient.GetContainerReference(dummyContainer)

	err = containerRef.Create(nil)
	if err != nil {
		log.Fatalf("Failed to create container: %s", err)
	}

	blobRef := containerRef.GetBlobReference("test.txt")

	err = blobRef.PutAppendBlob(nil)
	if err != nil {
		log.Fatalf("Failed to create append blob: %s", err)
	}

	dummyData := "deadbeef"
	err = blobRef.AppendBlock([]byte(dummyData), nil)
	if err != nil {
		log.Fatalf("Failed to write to append blob: %s", err)
	}

	log.Print("That's all folks!")
}
```

[This check](https://github.com/Azure/Azurite/blob/8c61bf42e8fb5d2d8fdff9f8e14f1bb9bf34ecde/src/blob/handlers/AppendBlobHandler.ts#L32-L37) seems to be the root cause.

I guess it works with `github.com/azure/azure-storage-blob-go` but I can't use it because it doesn't yet come with support for connection strings (see Azure/azure-storage-blob-go#211 and Azure/azure-storage-blob-go#244).

mihaitodor:
Due to a bug in azure-sdk-for-go, the Append Block API call fails because this legacy library always sends a non-zero Content-Length header. This fix relaxes this Content-Length check when the `--loose` flag is set.

Also adds a bunch of tests for AppendBlobHandler. Hope you don't mind having `ts-mockito` as a dependency. Testing this functionality without mocking the metadatastore and the extentstore doesn't look feasible.

Fixes #682.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,6 +8,7 @@ Blob:
 
 - Fix an issue that result of blobs enumeration cannot be parsed by Azure SDK for Go.
 - Fix an issue that set tier to leased blob not work properly.
+- Skip Content-Length check for Append Block if the `--loose` flag is set.
 
 ## 2020.12 Version 3.10.0
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -6175,6 +6175,15 @@
       "resolved": "https://registry.npmjs.org/triple-beam/-/triple-beam-1.3.0.tgz",
       "integrity": "sha512-XrHUvV5HpdLmIj4uVMxHggLbFSZYIn7HEWsqePZcI50pco+MPqJ50wMGY794X7AOOhxOBAjbkqfAbEe/QMp2Lw=="
     },
+    "ts-mockito": {
+      "version": "2.6.1",
+      "resolved": "https://repo1.uhc.com:443/artifactory/api/npm/npm-virtual/ts-mockito/-/ts-mockito-2.6.1.tgz",
+      "integrity": "sha1-vJ7iYZAzk05vrRxEVayltazjTnM=",
+      "dev": true,
+      "requires": {
+        "lodash": "^4.17.5"
+      }
+    },
     "ts-node": {
       "version": "7.0.1",
       "resolved": "https://registry.npmjs.org/ts-node/-/ts-node-7.0.1.tgz",
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "mocha": "^5.2.0",
     "prettier": "^1.16.4",
     "prettier-tslint": "^0.4.2",
+    "ts-mockito": "^2.6.1",
     "ts-node": "^7.0.1",
     "tslint": "^6.1.3",
     "typescript": "^3.1.4",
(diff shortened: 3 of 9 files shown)


blueww:
@XiaoningLiu 
Would you please help to look:
1. Is it OK to add ts-mockito as a dependency?
2. There are several "Insecure randomness" warning in code scan, are they OK?

XiaoningLiu (approved):

PR #710, opened 2021-02-28
Fix AppendBlock support for legacy azure-sdk-for-go

### LINKED ISSUE #702: azure-sdk-for-go AppendBlock operation resolves to AppendBlobHandler.create()

blob

Latest (v3.11.0)

DockerHub (`mcr.microsoft.com/azure-storage/azurite:3.11.0`)

Whatever the `mcr.microsoft.com/azure-storage/azurite:3.11.0` docker container is running

[`Blob.AppendBlock`](https://github.com/Azure/azure-sdk-for-go/blob/fd8f067989702b281ebb4177ddfbbba2651583e4/storage/appendblob.go#L70) gets resolved to [`AppendBlobHandler.create()`](https://github.com/Azure/Azurite/blob/456cbe74ca5c78b49e551b2659897ddf0d035b3c/src/blob/handlers/AppendBlobHandler.ts#L20) instead of [`AppendBlobHandler.appendBlock()`](https://github.com/Azure/Azurite/blob/456cbe74ca5c78b49e551b2659897ddf0d035b3c/src/blob/handlers/AppendBlobHandler.ts#L99). This is happening because the API specification [requires](https://github.com/Azure/Azurite/blob/456cbe74ca5c78b49e551b2659897ddf0d035b3c/swagger/blob-storage-2019-02-02.json#L7631-L7639) the `comp` query parameter for the `AppendBlob_AppendBlock` operation, while the `AppendBlob_Create` operation [requires](https://github.com/Azure/Azurite/blob/456cbe74ca5c78b49e551b2659897ddf0d035b3c/swagger/blob-storage-2019-02-02.json#L4285-L4300) the `x-ms-blob-type` header, which makes it equivalent to the `AppendBlob_Create` operation in [`dispatchMiddleware`](https://github.com/Azure/Azurite/blob/456cbe74ca5c78b49e551b2659897ddf0d035b3c/src/blob/generated/middleware/dispatch.middleware.ts#L47) (keys 59 and 60 return the same `conditionsMet`) because both of them contain the `Content-Length` header.

Would it be possible to make the `x-ms-blob-type` header mandatory for the `AppendBlob_AppendBlock` operation? This will probably require doing the same for `AppendBlob_AppendBlockFromUrl`, so it won't end up being confused with `AppendBlob_AppendBlock`. I'm happy to submit a PR for this if you can provide the steps to re-generate the code from the swagger definition.

Run the following code:

```go
package main

import (
	"io/ioutil"
	"log"
	"time"

	"github.com/Azure/azure-sdk-for-go/storage"
	"github.com/google/uuid"
	"github.com/ory/dockertest/v3"
	"github.com/ory/dockertest/v3/docker"
)

// RunAzurite starts an azurite container
func RunAzurite(pool *dockertest.Pool) (*dockertest.Resource, error) {
	opts := dockertest.RunOptions{
		Repository: "mcr.microsoft.com/azure-storage/azurite",
		Tag:        "3.11.0",
		PortBindings: map[docker.Port][]docker.PortBinding{
			"10000/tcp": {{HostIP: "0.0.0.0", HostPort: "10000"}},
		},
		Cmd: []string{"azurite", "--blobHost", "0.0.0.0", "--loose"},
	}
	azurite, err := pool.RunWithOptions(&opts)
	if err != nil {
		return nil, err
	}
	if eerr := azurite.Expire(10); eerr != nil {
		return nil, eerr
	}
	pool.MaxWait = 10 * time.Second
	rerr := pool.Retry(func() error {
		client, eerr := storage.NewEmulatorClient()
		if eerr != nil {
			return eerr
		}
		s := client.GetBlobService()
		c := s.GetContainerReference("cont")
		if _, err = c.Exists(); err != nil {
			return err
		}
		return nil
	})
	return azurite, rerr
}

func main() {
	pool, err := dockertest.NewPool("")
	if err != nil {
		log.Fatalf("Failed to create dockertest pool: %s", err)
	}

	azurite, err := RunAzurite(pool)
	defer pool.Purge(azurite)

	if err != nil {
		log.Fatalf("Failed to start Azurite: %s", err)
	}

	client, err := storage.NewEmulatorClient()
	if err != nil {
		log.Fatalf("Failed to create storage client: %s", err)
	}

	blobClient := client.GetBlobService()
	dummyContainer := uuid.New().String()
	containerRef := blobClient.GetContainerReference(dummyContainer)

	err = containerRef.Create(nil)
	if err != nil {
		log.Fatalf("Failed to create container: %s", err)
	}

	dummyBlob := uuid.New().String()
	blobRef := containerRef.GetBlobReference(dummyBlob)

	err = blobRef.PutAppendBlob(nil)
	if err != nil {
		log.Fatalf("Failed to create append blob: %s", err)
	}

	dummyData := uuid.New().String()
	err = blobRef.AppendBlock([]byte(dummyData), nil)
	if err != nil {
		log.Fatalf("Failed to write to append blob: %s", err)
	}

	blob, err := blobRef.Get(nil)
	if err != nil {
		log.Fatalf("Failed to get blob: %s", err)
	}

	data, err := ioutil.ReadAll(blob)
	if err != nil {
		log.Fatalf("Failed to read blob: %s", err)
	}

	if string(data) != dummyData {
		log.Fatalf("Expected %q but received %q instead", dummyData, string(data))
	}

	log.Print("That's all folks!")
}
```

Sadly, no... I would very much hope to see this fixed, if possible. Upgrading to the new azure-storage-blob-go API is not feasible for now.

mihaitodor:
Fixes #702.

Unfortunately, I don't see any straightforward way of adding tests for this since `tests/blob/apis/appendblob.test.ts` uses an auto-generated client, which behaves correctly, unlike github.com/Azure/azure-sdk-for-go/storage. Hope that's acceptable, since the extra logic is active only when the `--loose` flag is set.

LE: I'm not sure why the `Component Governance Component Detection` check has failed. Please let me know if there's anything that I need to change in my code for this to succeed.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,8 @@
 
 ## Upcoming Release
 
+- Resolve `AppendBlock` requests coming from github.com/Azure/azure-sdk-for-go/storage to the `appendBlock` method if the `--loose` flag is set.
+
 ## 2021.2 Version 3.11.0
 
 - Bump up Azure Storage service API version to 2020-06-12.
--- a/src/blob/handlers/AppendBlobHandler.ts
+++ b/src/blob/handlers/AppendBlobHandler.ts
@@ -29,6 +29,24 @@ export default class AppendBlobHandler extends BaseHandler
     const date = blobCtx.startTime!;
     const etag = newEtag();
 
+    // Hack: Forward requests with query containing `comp=appendblock` to the
+    // `appendBlock` method when in loose mode.
+    // This is required to support the AppendBlock API call via the legacy
+    // github.com/Azure/azure-sdk-for-go/storage client library.
+    // Details here: https://github.com/Azure/Azurite/issues/702
+    if (
+      context.request!.getQuery("comp") === "appendblock" &&
+      contentLength !== 0 &&
+      this.loose
+    ) {
+      return this.appendBlock(
+        context.request!.getBodyStream(),
+        contentLength,
+        options,
+        context
+      );
+    }
+
     if (contentLength !== 0 && !this.loose) {
       throw StorageErrorFactory.getInvalidOperation(
         blobCtx.contextId!,


blueww:
@mihaitodor 
As this is Go legacy SDK issue, it might not be so good to change Azurite for it.
Would you like try to raise a PR for Go SDK repo to fix the issue?

mihaitodor:
@blueww Thanks! I replied in #702.

mihaitodor:
Closing this, since the new azblob implementation should suit.

PR #848, opened 2021-06-24
Adds @Element suffix to odata.metadata for entity operation.

### LINKED ISSUE #133: Incorrect OData Metadata for single entity response

edwin-huber:
Fixes #133 

- Adds @Element suffix to odata.metadata for entity operation.
- Updated ChangeLog.md
- Added Test


src/table/utils/utils.ts
```diff
@@ -201,7 +201,7 @@ export function getOdataAnnotations(
     }
 
     return {
-      odatametadata: `${urlPrefixEndWithSlash}$metadata#${table}`,
+      odatametadata: `${urlPrefixEndWithSlash}$metadata#${table}/@Element`,
```
XiaoningLiu, line 204: Method `getOdataAnnotations` is used to generate annotations for Table operations and Entity operations. We should differentiate for Table operations.

XiaoningLiu (changes requested):

src/table/utils/utils.ts
edwin-huber: Adding tests to cover both cases.

XiaoningLiu (approved):

src/table/utils/utils.ts
edwin-huber: During testing, have uncovered deviations from current service responses... addressing these here as well

PR #927, opened 2021-07-18
fixes #722, query result is now complete using pagination

### LINKED ISSUE #722: The query result is incomplete when use Storage Table query entities by page operation in Azurite
**Error Description:**
The query result is incomplete when use Storage Table query entities by page operation in Azurite.
For example, in following demo, the count of table entities is 20 but actual query result is just 12. The running result is shown as following:
image

**To Reproduce:**
The demo code is shown as following:
```
import { odata, TableClient, TableEntity } from "@azure/data-tables"
import { assert } from "chai";

async function main() {
  const client = TableClient.fromConnectionString("<Azurite-https-connectionString>", "test");
  await client.create();
  const listPartitionKey = "listEntitiesTest";

  for (let i = 0; i < 20; i++) {
    await client.createEntity({
      partitionKey: listPartitionKey,
      rowKey: `${i}`,
      foo: "testEntity"
    });
  }

  const totalItems = 20;
  const maxPageSize = 5;
  const entities = client.listEntities<TableEntity<String>>({
    queryOptions: { filter: odata`PartitionKey eq ${listPartitionKey}` }
  });
  let all: TableEntity<String>[] = [];
  for await (const entity of entities.byPage({
    maxPageSize
  })) {
    all = [...all, ...entity];
  }
  assert.lengthOf(all, totalItems);
}

main().catch((err) => {
  console.log(err);
});
```

**Error Track:**
It will receive the maximal row key as the `nextRowKey` in every query by page. The code is shown [here](https://github.com/Azure/Azurite/blob/table/src/table/persistence/LokiTableMetadataStore.ts#L474)
Screenshot 2021-03-31 155658
then, the next query will use `nextRowKey` received from last query as filter to query entities whose row key is more than `nextRowKey`. the process code is shown [here](https://github.com/Azure/Azurite/blob/table/src/table/persistence/LokiTableMetadataStore.ts#L441)
image
Therefore, it can't query entity whose row key is less than `nextRowKey` in next query.

**Expected Behavior:**
The query result is complete when use Storage Table query entities by page operation in Azurite. 

@jongio for notification.

edwin-huber:
added test case to cover pagination number of results, and the distinct results returned by the query.
Also tested using different numbers of entities.  
Fixes #722

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -12,6 +12,7 @@ Table:
 - Added support for PATCH Verb in Table Batch Operations / Entity Group Transactions.
 - Added /@Element to the odata.metadata response.
 - Allowed use of empty string for partitionKey and rowKey on InsertEntity.
+- Correctly returning the results of paginated queries
 
 Blob:
 
--- a/src/table/persistence/LokiTableMetadataStore.ts
+++ b/src/table/persistence/LokiTableMetadataStore.ts
@@ -433,20 +433,27 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
       throw StorageErrorFactory.getQueryConditionInvalid(context);
     }
 
-    const segmentFilter = {} as any;
-    if (nextPartitionKey) {
-      segmentFilter.PartitionKey = { $gte: nextPartitionKey };
-    }
-    if (nextRowKey) {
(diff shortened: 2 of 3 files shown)


XiaoningLiu (approved):
LGTM

PR #934, opened 2021-07-20
Table Service GetProperties

edwin-huber:
added get table service properties and test

**NOTE: required change to generated specification artifacts and dispatch middleware**
Needed to remove the query parameter from the dispatcher for "timeout" as this is not always submitted by the SDKs.
Had to add a check for the case of service properties on the dispatcher which checks the URL path.

This is the first piece of work required to also support setting the table service properties.


XiaoningLiu (changes requested):
Please make following updates to `tableStorageContext.middleware.ts` line 85 then we will not need any change to the generated files.

image

justinmchase (approved):

XiaoningLiu (approved):

PR #972, opened 2021-07-28
Corrected tokenization of queries in table storage

### LINKED ISSUE #970: Table storage query returning unexpected rows
Table

3.14.0

DockerHub

QueryAsync is returning rows, even though none should match the query.

This works as expected with 3.13.1 and with real Azure Storage Tables.

```csharp
using Azure;
using Azure.Data.Tables;
using System;
using System.Threading.Tasks;

namespace TableStorageQueryBug
{
    class Program
    {
        static async Task Main(string[] args)
        {
            var client = new TableServiceClient("UseDevelopmentStorage=true");

            client.CreateTable("Testing123");

            var table = client.GetTableClient("Testing123");

            var oldestTimestamp = DateTimeOffset.UtcNow.AddDays(1);
            for (int i = 0; i < 10; i++)
            table.AddEntity(new Entity 
            { 
                PartitionKey = "A",
                RowKey = Guid.NewGuid().ToString(),
                Number = i
            });

            var pages = table.QueryAsync<Entity>(x => x.Number > 11 && x.Timestamp < oldestTimestamp).AsPages();

            int count = 0;
            await foreach (var page in pages)
            {
                foreach (var entity in page.Values)
                {
                    Console.WriteLine(entity.Number.ToString());
                    count++;
                }
            }

            Console.WriteLine("Count: " + count.ToString());

            client.DeleteTable("Testing123");

        }
    }

    public class Entity : ITableEntity
    {
        public string PartitionKey { get; set; } = default!;
        public string RowKey { get; set; } = default!;
        public DateTimeOffset? Timestamp { get; set; }
        public ETag ETag { get; set; }

        public int Number { get; set; }
    }
}
```

This should return no rows, as none of the 'Entity' objects have a Number property greater than 11.

However it gives this output:

```text
6
5
4
9
8
7
3
2
Count: 8
```

Interestingly, if the second part of the condition (`&& x.Timestamp < oldestTimestamp`) is removed, then it does return zero.

[debug.log](https://github.com/Azure/Azurite/files/6890827/debug.log)

Revert back to 3.13.1

edwin-huber:
fixes #970 

Fixes an issue in the tokenization of queries which contain numbers and no spaces between number and parens in the query filter.
This occurs due to differences in some SDKs.

Added test to cover this change and test the behavior using differently formatted query strings.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,6 +8,7 @@ Table:
 
 - Correctly returning the results of paginated queries.
 - Added filter support for Query Tables operation.
+- Corrected tokenization of queries in table storage.
 
 Blob:
 
--- a/src/table/persistence/LokiTableMetadataStore.ts
+++ b/src/table/persistence/LokiTableMetadataStore.ts
@@ -750,6 +750,10 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
           inString = false;
         }
       } else if (query[i] === "(" || query[i] === ")") {
+        if (i !== 0 && query[i - 1].match(/\d/) !== null) {
+          // this is needed if query does not contain whitespace between number token and paren
+          appendToken();
+        }
         i--;
         appendToken();
         i++;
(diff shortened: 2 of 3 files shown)


flcdrg:
Just to point out that the original code that had exhibited this problem didn't have a hard-coded number in the query - it was referencing an `int` variable. (I'd removed this for the bug repro)

edwin-huber:
@flcdrg thanks... the debug log that you provided was very helpful in identifying the root cause.
It shows us what is in the request when it reaches the API, rather than the code used in the client SDK.  
Whilst we can and do repro / test with C#, we write the tests in Typescript for our pipelines.
We also have the option to submit rest requests directly in these tests.

flcdrg:
Great to hear. Thanks for the quick response! 🙂

XiaoningLiu (approved):

PR #980, opened 2021-08-02
Table set service properties

edwin-huber:
Adds support for setServiceProperties

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -7,6 +7,7 @@
 Table:
 
 - Added getServiceProperties response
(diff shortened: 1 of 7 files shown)


edwin-huber:
resolves #961

larskeller (approved):

edwin-huber:
@XiaoningLiu discussed with @larskeller and asked him to test the approve function.  
Seems this is not locked down, however Lars is in the Microsoft org.

XiaoningLiu (approved):

PR #1017, opened 2021-08-24
Batch transaction error handling

RunyaoFan:
Creates batch transaction rollback mechanism in case of sub-request failure

### DIFF
--- a/src/table/batch/TableBatchOrchestrator.ts
+++ b/src/table/batch/TableBatchOrchestrator.ts
@@ -14,6 +14,8 @@ import {
   TableQueryEntitiesWithPartitionAndRowKeyOptionalParams
 } from "../generated/artifacts/models";
 import BatchTableQueryEntitiesOptionalParams from "./BatchTableQueryEntitiesOptionalParams";
+import ITableMetadataStore from "../persistence/ITableMetadataStore";
+import { v4 as uuidv4 } from 'uuid';
 
 /**
  * Currently there is a single distinct and concrete implementation of batch /
@@ -50,7 +52,8 @@ export default class TableBatchOrchestrator {
    * @memberof TableBatchManager
    */
   public async processBatchRequestAndSerializeResponse(
-    batchRequestBody: string
+    batchRequestBody: string,
+    metadataStore : ITableMetadataStore
   ): Promise<string> {
     this.batchOperations = this.serialization.deserializeBatchRequest(
       batchRequestBody
@@ -62,7 +65,7 @@ export default class TableBatchOrchestrator {
         this.context.xMsRequestID
       );
     } else {
-      await this.submitRequestsToHandlers();
+      await this.submitRequestsToHandlers(metadataStore, this.context);
     }
     return this.serializeResponses();
   }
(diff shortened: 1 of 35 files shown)


edwin-huber:
Copying a partition to validate the changes against it is a nice idea, but I am concerned that this might generate excessive load when Azurite has been running for a while with some of our SDKs.  
We need to test this with Durable Functions and logic apps, which make heavy use of batch requests, as well as it's behavior with larger partitions.

edwin-huber:
Currently working on some perf improvements for this PR as agreed with @XiaoningLiu

src/table/batch/TableBatchOrchestrator.ts
```diff
@@ -75,34 +78,99 @@ export default class TableBatchOrchestrator {
...
       const request: BatchRequest = new BatchRequest(operation);
       this.requests.push(request);
     });
 
     let contentID = 1; // ToDO: validate contentID starts at 1 for batch
     if (this.requests.length > 0) {
+      // TO DO: the way used to get accountName, tableName, and partition key
+      // may not work if incoming request is not compliant with the right format
+      const reqUrl = this.requests[0].getUrl();
+      const accountName = reqUrl.split("/")[3];
```
XiaoningLiu, line 92: Azurite supports 2 kinds of URL pattern.
- Path style. Storage account name is in URL path. Like "http:/localhost:10000/accountname"
- Production style. Storage account name is in first segment of URL endpoint. Like "http:://account.suffix/

XiaoningLiu: Refer to https://github.com/azure/azurite/blob/1ab42a545be0eda8a9d2d10e967e24e327374278/src/table/middleware/tableStorageContext.middleware.ts#L74 for existing method to extract account name and table name.

src/table/generated/handlers/ITableHandler.ts
```diff
@@ -17,13 +17,13 @@ export default interface ITableHandler {
...
-  queryEntities(table: string, options: Models.TableQueryEntitiesOptionalParams, context: Context): Promise<Models.TableQueryEntitiesResponse>;
-  queryEntitiesWithPartitionAndRowKey(table: string, partitionKey: string, rowKey: string, options: Models.TableQueryEntitiesWithPartitionAndRowKeyOptionalParams, context: Context): Promise<Models.TableQueryEntitiesWithPartitionAndRowKeyResponse>;
-  updateEntity(table: string, partitionKey: string, rowKey: string, options: Models.TableUpdateEntityOptionalParams, context: Context): Promise<Models.TableUpdateEntityResponse>;
-  mergeEntity(table: string, partitionKey: string, rowKey: string, options: Models.TableMergeEntityOptionalParams, context: Context): Promise<Models.TableMergeEntityResponse>;
-  deleteEntity(table: string, partitionKey: string, rowKey: string, ifMatch: string, options: Models.TableDeleteEntityOptionalParams, context: Context): Promise<Models.TableDeleteEntityResponse>;
+  queryEntities(table: string, options: Models.TableQueryEntitiesOptionalParams, context: Context, batchID?: string): Promise<Models.TableQueryEntitiesResponse>;
```
XiaoningLiu, line 20: We should not manually change files under "generated folder". As we have done in another PR, batchId is included in TableContext.

Issue #391, opened 2020-02-05
VSCode extension should create and use .azurite folder in current working directory when no location is specified


hannesne:
### Which service(blob, file, queue, table) does this issue concern?
NA
### Which version of the Azurite was used?
v3
### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
NA
### What's the Node.js version?
NA
### What problem was encountered?
The default lcoaiton for azurite to create it's files is the current working directory when invoked from the vscode extension. This creates a bunch of files and folders, excessively polluting the workspace. It would be better to create a .azurite directory in the current directory and put all the required files in there. This folder can then be added to .gitignore and other ignore files.
It is not convenient to have to set the location to a fixed directory to reference a subfolder in the current working directory.

### Steps to reproduce the issue?
NA
### Have you found a mitigation/solution?
NA

XiaoningLiu:
Hi @hannesne That's a great feedback! We will consider this in next major release as a breaking change. Before that, please define Azurite location in Visual Studio settings.

![image](https://user-images.githubusercontent.com/6889160/74102575-bf432580-4b7f-11ea-8dc8-395dea76b867.png)

JustinGrote:
@XiaoningLiu can you enable the use of variables such as ${workspaceFolder} for the Azure: Location setting? Currently I have to hard code it absolutely for each workspace.

empz:
It's actually really weird that the default behaviour is to output in the current working directory, but then there's no easy way to append a path to it without hardcoding the full directory path.

My suggestion is to make it accept both absolute and relative paths. And yes, make the default to be `.azurite` making the extension to put all files inside the current working directory / .azurite folder as suggested by the issue opener.

XiaoningLiu:
That's a valid ask and tracked on our radar. In the same time, feel free to send out PR which will speed up.

DaRosenberg:
Two more things that should perhaps be addressed as part of this:

- support expansion of environment variables in configured path, so that we could use e.g. `$TMPDIR/.azurite` on macOS
- the configured directory should be created if it does not already exist

blueww:
The feature request is also mentioned in https://github.com/Azure/Azurite/issues/962

dpolivy:
I took a stab at pulling together a PR to address this, as it was bugging me to pollute my git working directory with so many files. Hopefully this is acceptable and can be merged soon, but certainly open to feedback.

TroyWitthoeft:
[...]

PR #1061, opened 2021-09-24
Store files in '.azurite/' subdirectory of cwd by default, and ensure location exists

dpolivy:
This is a change to the default directory used to store Azurite files to utilize an '.azurite' folder in the current working directory. This allows for consolidation of files without polluting the working directory, and makes it easier to `.gitignore` the files.

Additionally, this change now ensures the directory exists on startup, creating it if it does not.

Partially Fixes https://github.com/Azure/Azurite/issues/391 -- I was not able to add the ability to use relative paths nor workspace variables as requested, but the main requests in the issue are addressed with this PR.

I am not sure how you'd like to manage this change as it is potentially breaking. I'm also not sure how/where to build appropriate tests for this.


XiaoningLiu:
Thanks @dpolivy! It helps a lot!

As you mentioned, it's a breaking change. How about split into 2 separate PRs:

- [Non-Breaking ] Support creating unexisted path, and support relative path in VSC by location parameter. This allows us specifying a directory in the configuration.
- [Breaking] Change the default location value for azurite. I'd like leave this breaking change in next major release.

dpolivy:
@XiaoningLiu Thanks for taking a look. I didn't really resolve the relative/absolute path in the location setting issue, though I suppose that could be done with [path.isAbsolutePath](https://nodejs.org/api/path.html#path_path_isabsolute_path).

I don't have a ton of time to devote to this but could probably do one more pass to add the above and split this into 2 PRs. Before I do that, it'd be great if you could review and advise if there's any other changes or additions you'd like to see.

I also didn't see any easy way to resolve VSCode variables in the settings; if you know of a way to do this easily, let me know and I can look at including it too.

dpolivy:
@XiaoningLiu Now that the other PR was merged, I cleaned up this one which sets the default directory to `.azurite`. I know you wanted to save this for a breaking change release, but let me know if there's anything else that needs to be done now. Or, I can just hand this over to you to manage going forward as you see fit.

TroyWitthoeft:
@XiaoningLiu - Any update on this review?  I know quite a few users who would really appreciate the default setting of azurite to be consolidating it's files into a `./azurite` subfolder.

DaRosenberg:
Come on now maintainers, this PR has been open since September last year, it's a small simple change that addresses a real issue faced by users (our team included). It's frustrating to see people donate their time and effort to fix a real issue, only to have their contribution completely ignored. Not good open source culture IMO - you can do better MS. Ping @XiaoningLiu since you're the current assignee.

XiaoningLiu:
For any PRs including breaking changes, we will wait for next major release by default, i.e Azurite v4.

For example, following scenario should work under same major version releases:
1. User lunches Azurite v3.18.0 in VSC under project1. Azurite creates database files under directory c:/project1
2. User upgrades Azurite to v3.19.0 in VSC under same project1.
3. User lunches Azurite 3.19.0. We will expect Azurite 3.19.0 loads databases from 3.18.0 by default.

Obviously, the PR is still breaking. In step#3, Azurite 3.19.0 will try to load and create databases from c:/project1/.azurite and ignore files under c:/project1. We will treat it as a breaking change. It breakes the user experiences and brings additional manual changes for someone upgrades Azurite version.

For quick merge, please consider making it non-breaking. For example, by adding a new parameter as flag to turn on the behavior. The default value of the parameter should be false in current major version, and in next major version, we can enable the parameter to true be default.

By the way, the PR should create .azurite folder if it doesn't exist.

JustinGrote:
So if this PR is updated to do the following:
1. Check if default azurite files exist and use those if present
2. If not, default to .azurite folder (or whatever is in settings)

You will accept the PR?

JustinGrote:
> @JustinGrote IMO a better order would be:
> 
> 1. If a location is set in settings, always use that
> 2. Otherwise if files exist in legacy default location, use that
> 3. Otherwise use new `.azurite` dir

Well yes, I meant to imply #1

TroyWitthoeft:
Ah, is this happening?   Yes!  🎉

JustinGrote:
@TroyWitthoeft  if @dpolivy doesn't pick it up, I might give it a go.

DaRosenberg:
@JustinGrote Would be awesome if you did! 🍻

One slight drawback of this approach though is that it would not be possible to use any other directory relative to the workspace than the default `.azurite` - any explicitly specified location would still be absolute.

The best non-breaking way I see to rectify that, would be to also add support for [VS Code variables](https://code.visualstudio.com/docs/editor/variables-reference#_predefined-variables) so that one could use `${workspaceFolder}` in the setting.

As @edwin-huber mentions above however, generically supporting resolution of all VS Code variables may not be possible in this context. But I don't see why we could not just add code to replace the tokens `${workspaceFolder}` and `${workspaceFolder:name}` with the corresponding workspace path provided to us by the API:
https://code.visualstudio.com/api/references/vscode-api#workspace

To support multi-root workspaces, we should use the `workspaceFolders` property and support named workspace folders as described here:
https://code.visualstudio.com/docs/editor/variables-reference#_variables-scoped-per-workspace-folder 

And mention in docs, settings hint etc. that this particular variable token is supported.

Another big bonus, at least from my point of view, would be also [supporting environment variables in the path](https://code.visualstudio.com/docs/editor/variables-reference#_environment-variables), which I think would also definitely qualify as non-breaking.

dpolivy:
Thanks for pushing on this, everyone. It's been a while since I worked on this. While I like the approach of trying to make this not a breaking change, from what I recall that's going to be much easier said than done due to the separation between the configuration/environment which defines the location path and the code that actually tries to load the database files to see if they exist.

I don't know that I'll have the ability to revisit this right away, so if someone is motivated to dig in, I'm happy to provide write access to my fork so you can update this PR.

> One slight drawback of this approach though is that it would not be possible to use any other directory relative to the workspace than the default `.azurite` - any explicitly specified location would still be absolute.

A prior PR (#1153) I submitted added support for relative paths to the location setting; this has been merged and should be in the code today.

DaRosenberg:
@dpolivy Dude.. 😅 I had no idea relative folder support had already been merged. That actually fixes our issues, for all intents and purposes. Thank you!

OOC though, do you know if there is any way to configure this ⬇️ in workspace settings?

image

We have a multi-root workspace, and having to select one of them every time Azurite starts is a bit of a pain... any way around it? Like an `azurite.workspaceName` setting or similar?

dpolivy:
> OOC though, do you know if there is any way to configure this ⬇️ in workspace settings?

It doesn't look like there's any way to avoid that currently. If there are multiple roots, it looks like it always pops that dialog: https://github.com/Azure/Azurite/blob/main/src/common/VSCEnvironment.ts#L51

potatoqualitee:
Any chance we can get some movement on this? 🙏🏼

blueww:
This is a breaking change.
Breaking is only allowed in major release.
So we will not take it until the next major release.

Xiaoning has given some suggestion to convert it to a none breaking change.
If it's not breaking, we can take it when pass review.

paulyuk:
Hi - I do not believe this is a breaking change and we need to fix the more dire issue that Azurite extension writes your data files to the workspace/repo where it is adding unwanted, possibly secure data to the repo.  Could we please fix this with higher priority?  Across Azure Functions I no longer recommend it, pending a fix to this.  Thank you!  pls ping if you have questions.

JustinGrote:
> Hi @paulyuk ,
> 
> We have a internal mail to follow up this, and you are also in the thread. Let's first discuss in that mail and make decision, then see what should we do for this PR.
> 
> Besides that, this Do be a breaking change. The customer who has Azurite run with some data saved, will find the data disappear after upgrade to a new Azurite version.

Why does it need to *disappear*? Why not just have a notice saying the path has changed and to migrate their data? Even better, have the dialog have a button to move the data automatically.

JustinGrote:
Just an information message dialog in the lower right that detects of the old file paths exists and notify to move them.

PR #1080, opened 2021-10-05
Fixes issue with incorrect results when using Boolean values in table query

### LINKED ISSUE #1051: bad query result when using a boolean clause with parenthesis

table

3.14.2

npm

14.17.6

Azurite does not give correct query results (it finds none) when a query contains a clause of this kind:
`(MyBooleanField eq true)`
On the contrary, the deprecated azure storage emulator and actual azure storage table service give results when using this clause.

However, this clause works (Notice the space before closing the parenthesis):
`(MyBooleanField eq true )`

- Have a table with several fields, at least one bool.
- Fill this table with a few entities.
- Open a storage explorer, or have a small application that queries the service (e.g. Azure Cosmos Table client library for .Net)
- Execute a query with a filtering clause on the boolean field. It should give results.
- Execute the same query with parenthesis around the boolean clause. It gives no result.
- Add a space before closing the parenthesis. It gives results.

Example anonymized log attached (I replace results when I got some by a filler text, line 15) : [example.log](https://github.com/Azure/Azurite/files/7178900/example.log)

Not for now, a lot of libraries do not let you manipulate the query string directly.

edwin-huber:
fixes #1051
Contains test case to validate change.
Corrects case for query conversion in metadata store if no space between Boolean value and parens.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,6 +11,10 @@ Blob:
 - Added more allowed value to list blob request "include" query parameter:'tags', 'versions', 'deletedwithversions', 'immutabilitypolicy', 'legalhold', 'permissions'
 - Added more allowed value to list container request "include" query parameter: 'deleted'
 
+Table:
+
+- Fixed issue with incorrect results returned when using boolean values in query.
+
 ## 2021.9 Version 3.14.2
 
 Blob:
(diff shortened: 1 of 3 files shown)


edwin-huber:
Also now fixes #1084 for int64 values in query

Issue #1142, opened 2021-11-01
Blob storage returns 403 when trying to ready from secondary on v3.14.2+


vikene:
### Which service(blob, file, queue, table) does this issue concern?
Blob
### Which version of the Azurite was used?
3.14.2
### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
NPM
### What's the Node.js version?
14.17.4
### What problem was encountered?
When we try to access secondary location of blob storage, it returns 403 - forbidden error. 
### Steps to reproduce the issue?
Attaching debug log and project to replicate this issue. 
Extract the project and give following commands in order to replicate this issue.
    1) Dotnet build 
    2) Dotnet run
[blobstorage.zip](https://github.com/Azure/Azurite/files/7454811/blobstorage.zip)

If possible, please provide the debug log using the -d parameter, replacing \<pathtodebuglog\> with an appropriate path for your OS, or review the instructions for docker containers:
Attaching debug log as a file.
[debug.log](https://github.com/Azure/Azurite/files/7454804/debug.log)

Please be sure to remove any PII or sensitive information before sharing!  
The debug log will log raw request headers and bodies, so that we can replay these against Azurite using REST and create tests to validate resolution.

### Have you found a mitigation/solution?
Fall back to 3.14.0

vigneashs:
Created a pull request for this issue - https://github.com/Azure/Azurite/pull/1145

blueww:
The issue has been fixed in Azurite 3.15.0.

PR #1145, opened 2021-11-01
Removes code to apprend -secondary flag. for issue #1142

vigneashs:
This PR removes code to append -secondary flag in authenticator. This URI change should be taken care by the client and not by the authenticator.

Skips test cases that require secondary URI. to be re-enabled after this bug https://github.com/Azure/azure-sdk-for-js/issues/13033 is fixed.

### PR Branch Destination
Master branch

###Test Cases
This PR skips three test cases as they are requiring a hard-coded secondary URL. This should be passed as an option to client and client should replace it with correct path. Since, this feature isn't there yet in @azure/storage due to bug -> https://github.com/Azure/azure-sdk-for-js/issues/13033, leaving it as a revisit item.

### DIFF
--- a/src/blob/authentication/BlobSharedKeyAuthenticator.ts
+++ b/src/blob/authentication/BlobSharedKeyAuthenticator.ts
@@ -69,7 +69,7 @@ export default class BlobSharedKeyAuthenticator implements IAuthenticator {
       this.getCanonicalizedResourceString(
         req,
         account,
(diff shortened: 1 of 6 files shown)


vigneashs:
For further reference on my change. https://docs.microsoft.com/en-us/rest/api/storageservices/authorize-with-shared-key#constructing-the-canonicalized-resource-string.

PR #1147, opened 2021-11-02
Ensure storage location exists and allow relative paths

dpolivy:
This commit is a follow-up to #1061 which only contains the two features below:

1. Ensure directories exist on startup, and create them if not.
2. Allow specifying relative paths in VSCode settings that are resolved based on the workspace folder.

I was not able to add any automated tests for this, but overall it seems to be working well in my manual testing.

PR #1153, opened 2021-11-03
Ensure storage location exists and allow relative paths

blueww:
This commit is a follow-up to #1061 which only contains the two features below:

Ensure directories exist on startup, and create them if not.
Allow specifying relative paths in VSCode settings that are resolved based on the workspace folder.
I was not able to add any automated tests for this, but overall it seems to be working well in my manual testing.

The PR has an origin PR https://github.com/Azure/Azurite/pull/1147, which was merged by mistake, and reverted. So raise again.


blueww:
@dpolivy 
I have re-raised the commit in PR https://github.com/Azure/Azurite/pull/1147 in this PR.
If you would like a better track for the PR progress, you can close this and open a new one.

src/common/VSCEnvironment.ts
```diff
@@ -64,10 +62,11 @@ export default class VSCEnvironment implements IEnvironment {
       } else {
         folder = workspace.workspaceFolders[0];
       }
-      location = folder.uri.fsPath;
+      location = resolve(folder.uri.fsPath, location ?? '');
```
blueww, line 65: This might be a breaking change for the existing user, who use VS extension and relative path.
Could we remove the part of change, and keep the original folder?

We can add the breaking change in  a major release.

src/queue/main.ts
```diff
@@ -36,11 +33,13 @@ async function main() {
   const env = new QueueEnvironment();
 
   const location = await env.location();
-  await accessAsync(location);
+  await ensureDir(location);
```
blueww, line 36: It looks like ensureDir() has been called several times on a folder . Like it has been called when get env.location().

dpolivy:
> @dpolivy I have re-raised the commit in PR #1147 in this PR. If you would like a better track for the PR progress, you can close this and open a new one.

@blueww I'm not sure I have the ability to close this, but I'm happy to open a new one if you like for better tracking. In the meantime, I will reply to your comments above and can address in a new PR if needed.

src/common/VSCEnvironment.ts
dpolivy: Well, I broke apart my existing PR as requested by @XiaoningLiu in https://github.com/Azure/Azurite/pull/1061#issuecomment-927235519 and this is supposed to be the non-breaking change. The current behavior of the location setting does not support a relative path at all -- any value entered in the field will be treated as an absolute path. This change just adjusts the behavior so that if you enter a relative path, it will actually be relative. If you enter an absolute path, it will continue to be absolute.

If you enter what you expect to be a relative path, but it is treated as absolute, I suppose this could be "breaking" but one would also expect it to have worked as a relative path so it could also be considered a bug fix :)

src/queue/main.ts
dpolivy: In this example, `env` is coming from `QueueEnvironment`, which does not call `ensureDir` at all.

I admit, I'm not 100% certain of all the instantiation cases for this so I may have been over-zealous in adding this -- I essentially looked for anywhere that you did `access/accessAsync` and updated that to do both. If I understand the code correctly, if someone runs just the Queue service directly, this would be the only code that would ensure the queue location exists?

src/common/VSCEnvironment.ts
XiaoningLiu: Currently, if someone enters a relative path into VSC extension location settings like "abc". It's actually relative to Visual Studio Code installatoin path. For example, "c:/xxx/visual studio code/abc". 

Strictly it's a breaking for someone really depends on above behavior - "Create "abc" under Visual Studio Code path". Will any?

Well, we may treat this a "bug" and "wrong usage".

Please help update the location parameter description in package.json and Readme.md, to mention "what will happen to a relative path in VSC".

src/queue/main.ts
XiaoningLiu: Checked again. It's true that you add ensureDir for before every existing access check.

XiaoningLiu (commented):
refer to comments above

src/queue/main.ts
dpolivy: @XiaoningLiu So is there any further action you'd like me to take here or is it OK to leave the `ensureDir` calls as-is?

src/common/VSCEnvironment.ts
dpolivy: @XiaoningLiu I pushed up some doc changes to clarify this.

dpolivy:
@XiaoningLiu @blueww Any further comments/changes needed here for this PR?

XiaoningLiu:
@dpolivy Can you please add a changelog entry and resolve the conflicts? PR is generally good to us.

dpolivy:
@XiaoningLiu I rebased the branch off main and added a ChangeLog entry. Let me know if there's anything else!

XiaoningLiu (approved):

Issue #1119, opened 2021-10-20
Wrong data retrieved from table for TableEntity with property of type double


bragma:
### Which service(blob, file, queue, table) does this issue concern?
table

### Which version of the Azurite was used?
3.14.2

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
DockerHub

### What's the Node.js version?
v14.17.6

### What problem was encountered?
Retrieving data with ExecuteQuerySegmentedAsync (C#) on a class inherited from TableEntity and double properties retrieves 0 as value instead of 1.

### Steps to reproduce the issue?
Please see the test repository [here](https://github.com/bragma/AzuriteTableTest). Start Azurite docker image and run the program. Data will be correctly written to the table (check with Storage Explorer), but query will retrieve value 0 instead of 1.
Notice that changing property type from double to int will instead return the correct result.

### Have you found a mitigation/solution?
No

edwin-huber:
Hi,
I have been testing your repro.
In the current main branch, this is working, as are the tests which I created using Azure data-tables package.
Could you please confirm?
Thanks!

edwin-huber:
I should also add that the WindowsAzure.Storage Package that you are using has been deprecated.
Please check out Azure.Data.Tables as a replacement.
https://github.com/Azure/azure-sdk-for-net/blob/Azure.Data.Tables_12.2.1/sdk/tables/Azure.Data.Tables/README.md

bragma:
> I should also add that the WindowsAzure.Storage Package that you are using has been deprecated.

I know, thanks for reminiding me. I'm using that package specifically because it's the one we are using in production and may take some time for us to migrate.

bragma:
> Hi, I have been testing your repro. In the current main branch, this is working, as are the tests which I created using Azure data-tables package. Could you please confirm? Thanks!

I'll try. As I specified in the original report, I've been using the image from dockerhub (docker on windows WSL). I'm not sure if this may make a difference. Maybe the problem is with the dockerhub image.
In any case, I'll try to run from main locally and let you know. Thanks!

edwin-huber:
Ah, yes, the DockerHub image is not in parity with the latest PRs.

bragma:
> Ah, yes, the DockerHub image is not in parity with the latest PRs.

I suggest to use imege tags properly instead of "latest"

edwin-huber:
> I suggest to use imege tags properly instead of "latest"

[...]

PR #1155, opened 2021-11-03
Fixes backwards compatability for Azure-Storage SDK double value props

edwin-huber:
Adds test cases for double values using Azure Data Tables.


edwin-huber:
Fixes #1119

Azure Data-Tables package uses a number type in the json when inserting entities, Azure-Storage SDK used a string.
Fix addresses this in the MetaData layer by converting the string representation of a double to a number type in the database.

XiaoningLiu (approved):

lamjay2704:
Smith

PR #1164, opened 2021-11-10
Fix the sharedkey validation fail on secondary location with IP style Uri

blueww:

To Fix issue: https://github.com/Azure/Azurite/issues/1142

### PR Branch Destination

- For Azurite V3, please send PR to `master` branch.

### Always Add Test Cases

### Add Change Log

### Development Guideline

### DIFF
--- a/src/blob/authentication/BlobSharedKeyAuthenticator.ts
+++ b/src/blob/authentication/BlobSharedKeyAuthenticator.ts
@@ -69,7 +69,7 @@ export default class BlobSharedKeyAuthenticator implements IAuthenticator {
       this.getCanonicalizedResourceString(
         req,
         account,
-        context.context.isSecondary ? blobContext.authenticationPath + "-secondary" : blobContext.authenticationPath
+        blobContext.authenticationPath
       );
 
     this.logger.info(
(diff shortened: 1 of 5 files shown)


XiaoningLiu (approved):

src/blob/authentication/BlobSharedKeyAuthenticator.ts
```diff
@@ -112,6 +112,75 @@ export default class BlobSharedKeyAuthenticator implements IAuthenticator {
       }
     }
 
+    if (context.context.isSecondary && blobContext.authenticationPath?.indexOf(account) === 1)
+    {
+        // JS/.net Track2 SDK will generate stringToSign from IP style Uri with "-secondary" in authenticationPath, so will also compare signature with this kind stringToSign
+        const stringToSign_secondary: string =
```
vigneashs, line 118: Just a suggestion. Can we split this implementation into two parts ? 
example: Let's just have an array variable on the top, stringToSign. Without .Join operation on line number 52.
Later we can re-use the same array here to do the join operation, as we are repeating more than 10 lines which are just going to be same irrespective of it being secondary or not.

src/table/authentication/TableSharedKeyAuthenticator.ts
```diff
@@ -104,6 +104,66 @@ export default class TableSharedKeyAuthenticator implements IAuthenticator {
       }
     }
 
+    if (context.context.isSecondary && tableContext.authenticationPath?.indexOf(account) === 1)
+    {
+      // JS/.net Track2 SDK will generate stringToSign from IP style Uri with "-secondary" in authenticationPath, so will also compare signature with this kind stringToSign
+      const stringToSign_secondary: string =
```
vigneashs, line 110: Same comments as above. We can have one header array, that can be re-used.

vigneashs (approved):
Approving with suggestions 👍

src/blob/authentication/BlobSharedKeyAuthenticator.ts
blueww: @vigneashs 

Thanks for the suggestion!
To avoid regression, I don't want to change the origin code too much. So I would prefer to keep origin code not change,  and only add logic for more stringtosign format.
Let me know if you have any concern.

src/table/authentication/TableSharedKeyAuthenticator.ts
blueww: same as above. 
To avoid regression, I would prefer not change original code too much, but only add new code to addtional stringToSign format.

PR #1200, opened 2021-11-30
Added support for building Linux binaries

vigneashs:

### PR Branch Destination
This feature is targeted for Azurite V3

### Test Cases
Added test cases for Linux binary executables.
Also added a job to test Linux binary in pipeline.

### Change Log
VFS and other projects are expected to have developer loops on systems other than windows. This feature adds support to azurite to release Linux binaries.


azure-pipelines.yml
```diff
@@ -526,3 +526,40 @@ jobs:
         workingDirectory: "./"
         displayName: "npm run test:exe"
         env: {}
+
+  - job: linuxTest
+    displayName: binary Test Linux
+    pool:
+      vmImage: "ubuntu-latest"
```
XiaoningLiu, line 533: Is "ubuntu" only enough? Can you help add other linux OS as well into the coverage?

scripts/buildLinux.js
```diff
@@ -0,0 +1,43 @@
...
+
+  function asyncRename(oldName, changedName) {
+    return new Promise(resolve => {
+      fs.rename(oldName, changedName, response => resolve(response));
+    });
+  }
+
+  await asyncRename(cacheLinux, newName);
+
+  const outputExe = path.resolve('./release/azuritelinux');
```
XiaoningLiu, line 29: variable name?

tests/linuxbinary.test.ts
```diff
@@ -0,0 +1,364 @@
...
+  createConnectionStringForTest, HOST, PORT, PROTOCOL
+} from './table/apis/table.entity.test.utils';
+import {
+  bodyToString, EMULATOR_ACCOUNT_KEY, EMULATOR_ACCOUNT_NAME, getUniqueName, overrideRequest,
+  restoreBuildRequestOptions
+} from './testutils';
+
+// server address used for testing. Note that Azurite.exe has 
+// server address of http://127.0.0.1:10000 and so on by default 
+// and we need to configure them when starting azurite.exe
```
XiaoningLiu, line 29: `azurite.exe` ?

XiaoningLiu (commented):
Comments added, please look. Major question is how many Linux distributions will be supported?

azure-pipelines.yml
vigneashs: Sure, I used "ubuntu-latest", Since it is the default Microsoft hosted agent. For any other distributions, we might have to create runners/ agents in Azure devops. I tried to create one but looks like I don't have permissions to create those hosted agents in your ADO workspace. 

let me know if you are fine with creating these custom pool runners with multiple Linux distros.

Following are the distros that would be the recommended set of test agents, I arrived at this list based on flavors and forks of major Linux distros. Eg: Debian, Ubuntu, Mint is counted as Ubuntu.
1) Ubuntu
2) CentOS
3) Fedora
4) openSUSE
5) RHEL

@ayeltsov let me know your opinion on this, on VFS side we will only be supporting Ubuntu for test and development. (Windows subsystem for Linux only with Ubuntu is supported)

XiaoningLiu: If Ubuntu is the only requirement from your side, it's fine to keep Ubuntu only here. 
@blueww can you help check if we can have mentioned platforms into the pipeline?

tests/linuxbinary.test.ts
XiaoningLiu: I don't think you are generating "azurite.exe" for Linux distribution. Please correct them.

azure-pipelines.yml
blueww: @XiaoningLiu 
From https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software, we can see currently Azure Pipelines agent pool offers following linux virtual machine images : ubuntu-20.04 and ubuntu-18.04

vigneashs: Added ubuntu-18.04. let me know if this looks good

XiaoningLiu (approved):

XiaoningLiu:
azp /run

XiaoningLiu:
image

@vigneashs The pipeline has an issue on duplicated test job naming. Can you update?

vigneashs:
Fixed duplicate name @XiaoningLiu

blueww:
@vigneashs 

The ci failed since  a missing reference, could you please update against latest main, and see if the issue can be resolved?
the error is: 

> tests/linuxbinary.test.ts(21,8): error TS2307: Cannot find module './table/apis/table.entity.test.utils' or its corresponding type declarations.

Issue #1229, opened 2021-12-18
Table service returns 400 not 412 for an unexpected weak etag


joelverhagen:
### Which service(blob, file, queue, table) does this issue concern?

table

### Which version of the Azurite was used?

3.15.0

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)

VS Code extension

### What's the Node.js version?

16.13.0

### What problem was encountered?

When a non-standard weak etag, e.g. `W/"wrong"`, is provided, HTTP 400 is returns instead of HTTP 412. Azure and Storage Emulator return 412 for this etag.

This is not a huge deal but I happened to have a test in my code that was testing a bad etag like this expecting it to be accepted to the system but rejected for never matching. Since this is very much an edge case, feel free to close this if you don't want a fix. I'd be happy to provide a fix and test.

### Steps to reproduce the issue?

```csharp
using Azure;
using Azure.Data.Tables;

var tableClient = new TableServiceClient("UseDevelopmentStorage=true");
var table = tableClient.GetTableClient("table0");
table.Delete();
table.Create();
table.AddEntity(new TableEntity("PK", "RK1"));

try
{
    table.UpdateEntity(new TableEntity("PK", "RK1"), new ETag("W/\"wrong\""));
}
catch (RequestFailedException ex)
{
    Console.WriteLine("Status: " + ex.Status);
}
```

```
2021-12-18T23:35:59.952Z 	 info: Azurite Blob service is starting on 127.0.0.1:10000
2021-12-18T23:35:59.954Z 	 info: AccountDataStore:init() Refresh accounts from environment variable AZURITE_ACCOUNTS with value undefined
2021-12-18T23:35:59.954Z 	 info: AccountDataStore:init() Fallback to default emulator account devstoreaccount1.
2021-12-18T23:35:59.993Z 	 info: BlobGCManager:start() Starting BlobGCManager. Set status to Initializing.
2021-12-18T23:35:59.994Z 	 info: BlobGCManager:start() Trigger mark and sweep loop. Set status to Running.
2021-12-18T23:35:59.994Z 	 info: BlobGCManager:markSweepLoop() Start next mark and sweep.
2021-12-18T23:35:59.994Z 	 info: BlobGCManager:markSweep() Get all extents.
2021-12-18T23:35:59.995Z 	 info: BlobGCManager:start() BlobGCManager successfully started.
[...]

PR #1231, opened 2021-12-19
Allow any weak etag even if it doesn't match the pattern (412 should happen)

joelverhagen:
Fix https://github.com/Azure/Azurite/issues/1229

This is pretty low priority, but I wanted to start with an easy bug fix as I ramp up on this repository. This is only my 3rd PR 😄.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,6 +8,10 @@ General:
 
 - Ensure the storage location exists, and allow relative paths in the VSCode extension settings that are resolved based on the workspace folder.
 
+Table:
+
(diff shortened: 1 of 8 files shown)


XiaoningLiu (commented):
Thanks Joel! Seems the change is only for Table. If so, can you help move the `checkEtagIsInvalidFormat` from `common` into `table/utils.ts`?

joelverhagen:
Good call. Done. There was another pre-existing table-only util that I moved too.

joelverhagen:
@XiaoningLiu - could this be merged? Or are there more changes I should do?

Issue #1214, opened 2021-12-13
Table query page continuation throws 500 on some non-ASCII partition key or row keys


joelverhagen:
### Which service(blob, file, queue, table) does this issue concern?

table

### Which version of the Azurite was used?

3.15.0

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)

VS Code

### What's the Node.js version?

16.13.0

### What problem was encountered?

When querying table entities with some non-ASCII characters (e.g. `𤭢` or `𐐷`), the query fails with HTTP 500 when the continuation token aligns with this character.

### Steps to reproduce the issue?

```csharp
using Azure.Data.Tables;

var tableClient = new TableServiceClient("UseDevelopmentStorage=true");
var table = tableClient.GetTableClient("table0");
table.Delete();
table.Create();
table.UpsertEntity(new TableEntity("𤭢PK1", "𐐷RK1"));
table.UpsertEntity(new TableEntity("𤭢PK2", "𐐷RK2"));
foreach (var item in table.Query<TableEntity>(maxPerPage: 1))
{
    Console.WriteLine(item.PartitionKey + "/" + item.RowKey);
}
```

Logs

```
2021-12-13T18:29:18.069Z 	 info: Azurite Table service is starting on 127.0.0.1:10002
2021-12-13T18:29:18.071Z 	 info: AccountDataStore:init() Refresh accounts from environment variable AZURITE_ACCOUNTS with value undefined
2021-12-13T18:29:18.072Z 	 info: AccountDataStore:init() Fallback to default emulator account devstoreaccount1.
2021-12-13T18:29:18.100Z 	 info: Azurite Table service successfully listens on http://127.0.0.1:10002
[...]

PR #1235, opened 2021-12-19
Encode next partition key and row key using base64 to avoid Unicode issues

joelverhagen:
Fix https://github.com/Azure/Azurite/issues/1214


XiaoningLiu (approved):

XiaoningLiu:
Thanks for the contribution!

PR #1358, opened 2022-02-24
Table Query Filter validation

### LINKED ISSUE #1185: Emulated table service does not return RestError for invalid filter option
Table
3.14.3
npm
4.14.7
The  table service does not return with a RestError for invalid queries.
Run the following code in Node:
```js
const { TableClient } = require("@azure/data-tables");
const connectionString = "<DEV_CONNECTION_STRING>";
(async () => {
    const client = TableClient.fromConnectionString(connectionString, "test01", { allowInsecureConnection: true });
    try {
        await client.createTable();
        const result = await client.listEntities({ queryOptions: { filter: "1" } }).byPage();
        console.log(result.next(), null, 2);
    } catch (error) {
        console.error(error);
    }
})();
```

If the connection string used is for a dev account (emulator), the `console.log` statement is run, which is unexpected.

If you swap the connection string for a real Azure table service, the `console.error` statement is run, printing out the message "The $filter expression must evaluate to a single boolean value."

No

edwin-huber:
Adds Table Query Filter validation and appropriate tests.
Corrected table entity properties used in tests.
Responds with Status Code 400 on invalid query filters.

Fixes #1185


edwin-huber:
@XiaoningLiu Failures in CI are due to DataTables update including BigInt which is not available in Node 8 (which is in all the failing test cases).
Pending investigation and decision on how we resolve.

XiaoningLiu:
Will rerun after removal of node 8 PR merged.

src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -18,6 +18,17 @@ interface IServiceAdditionalProperties {
 export type ServicePropertiesModel = Models.TableServiceProperties &
   IServiceAdditionalProperties;
 
+// used by the query filter checking logic
+type tokenTuple = [string, tokenType];
+enum tokenType {
```
XiaoningLiu, line 23: We should have type and enum starting with upper case, like "TokenTuple" and "TokenType".

src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -18,6 +18,17 @@ interface IServiceAdditionalProperties {
...
+  unknown,
```
XiaoningLiu, line 24: enum values should start with upper case as well to align with Azurite coding practice

edwin-huber:
> Will rerun after removal of node 8 PR merged.

Node 8 PR is merged, but the old failures are not removed.

src/table/persistence/LokiTableMetadataStore.ts
edwin-huber: Casing changed on token type and enum.

src/table/persistence/LokiTableMetadataStore.ts
edwin-huber: Casing changed.

XiaoningLiu (approved):

Issue #1234, opened 2021-12-19
Empty partition key causes errors in table batch operations


joelverhagen:
### Which service(blob, file, queue, table) does this issue concern?

table

### Which version of the Azurite was used?

Locally running version of #1157

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)

Git clone

### What's the Node.js version?

16.13.0

### What problem was encountered?

When a table batch action (e.g. update-replace) contains an empty string partition key, an exception is thrown.

```
Unhandled exception. Azure.Data.Tables.TableTransactionFailedException: The specified table name is empty.
RequestId:40ec2c98-a25d-4a62-8130-e03b2bb3d7dd
Time:2021-12-19T07:30:17.983Z
Status: 400 (Bad Request)
ErrorCode: TableNameEmpty

Content:
{"odata.error":{"code":"TableNameEmpty","message":{"lang":"en-US","value":"The specified table name is empty.\nRequestId:40ec2c98-a25d-4a62-8130-e03b2bb3d7dd\nTime:2021-12-19T07:30:17.983Z"}}}

Headers:
Content-ID: REDACTED
DataServiceVersion: 3.0;
Content-Type: application/json;odata=minimalmetadata;charset=utf-8

   at Azure.Data.Tables.TableRestClient.SendBatchRequest(HttpMessage message, CancellationToken cancellationToken)
   at Azure.Data.Tables.TableClient.SubmitTransactionInternalAsync(IEnumerable`1 transactionalBatch, Guid batchId, Guid changesetId, Boolean async, CancellationToken cancellationToken)
   at Azure.Core.Pipeline.TaskExtensions.EnsureCompleted[T](Task`1 task)
   at Azure.Data.Tables.TableClient.SubmitTransaction(IEnumerable`1 transactionActions, CancellationToken cancellationToken)
   at Program.<Main>$(String[] args) in C:\Users\joelv\Desktop\AzuriteSample\AzuriteSample\Program.cs:line 9
```

### Steps to reproduce the issue?

```csharp
using Azure.Data.Tables;

var tableClient = new TableServiceClient("UseDevelopmentStorage=true");
var table = tableClient.GetTableClient("table0");
table.Delete();
table.Create();
table.AddEntity(new TableEntity("", "RK1"));
table.SubmitTransaction(new[]
{
    new TableTransactionAction(TableTransactionActionType.UpdateReplace, new TableEntity("", "RK1")),
});
```

```
[...]

PR #1364, opened 2022-02-28
Fix batch handling of empty partition key and empty row key

joelverhagen:
Resolve https://github.com/Azure/Azurite/issues/1234.

### DIFF
--- a/src/table/handlers/TableHandler.ts
+++ b/src/table/handlers/TableHandler.ts
@@ -196,10 +196,10 @@ export default class TableHandler extends BaseHandler implements ITableHandler {
       eTag: newTableEntityEtag(context.startTime!)
     };
 
-    let normalizedEntity;
+    let nomarlizedEntity;
     try {
-      normalizedEntity = new NormalizedEntity(entity);
-      normalizedEntity.normalize();
+      nomarlizedEntity = new NormalizedEntity(entity);
+      nomarlizedEntity.normalize();
     } catch (e: any) {
       this.logger.error(
         `TableHandler:insertEntity() ${e.name} ${JSON.stringify(e.stack)}`,
@@ -254,7 +254,7 @@ export default class TableHandler extends BaseHandler implements ITableHandler {
       // }
 
       // response.body = new BufferStream(Buffer.from(JSON.stringify(body)));
-      const rawResponse = normalizedEntity.toResponseString(accept, body);
+      const rawResponse = nomarlizedEntity.toResponseString(accept, body);
       this.logger.debug(
         `TableHandler:insertEntity() Raw response string is ${JSON.stringify(
           rawResponse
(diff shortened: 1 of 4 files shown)


joelverhagen:
@edwin-huber, I took a stab at https://github.com/Azure/Azurite/issues/1234. I saw you were assigned so I hope you haven't started working on it. If so, feel free to repurpose/toss out this PR 😃.

src/table/handlers/TableHandler.ts
```diff
@@ -359,16 +365,21 @@ export default class TableHandler extends BaseHandler implements ITableHandler {
 
   public async mergeEntity(
     _table: string,
-    partitionKey: string,
+    partitionKey: string | undefined,
```
edwin-huber, line 368: @XiaoningLiu extending the parameter in the function signature with undefined is valid based on the "Duck Typing" in Typescript.
Are you happy with this?

edwin-huber:
Thanks for the submission!
I like the changes; I have one small suggestion and would like @XiaoningLiu to weigh in on the Handler Method signature.

src/table/handlers/TableHandler.ts
joelverhagen: I think empty string is not a good indicator of missing value. An empty string is actually a valid partition key and `undefined` means the value should be picked from the context if possible.

XiaoningLiu (approved):

XiaoningLiu:
@edwin-huber Do you have other comments before merging this PR?

edwin-huber (changes requested):
Please collect all checks for values into a single function.
This will help us when we update validation logic in the future.

```
partitionKey = partitionKey !== undefined ? partitionKey : tableContext.partitionKey!;
```

src/table/handlers/TableHandler.ts
```diff
@@ -275,16 +275,22 @@ export default class TableHandler extends BaseHandler implements ITableHandler {
...
     const table = this.getAndCheckTableName(tableContext);
-    partitionKey = partitionKey || this.getAndCheckPartitionKey(tableContext);
-    rowKey = rowKey || this.getAndCheckRowKey(tableContext);
+
+    partitionKey = partitionKey !== undefined ? partitionKey : tableContext.partitionKey!; // Get partitionKey from context
+    rowKey = rowKey !== undefined ? rowKey : tableContext.rowKey!; // Get rowKey from context
+
+    if (partitionKey === undefined || rowKey === undefined) {
+      throw StorageErrorFactory.getPropertiesNeedValue(context);
+    }
```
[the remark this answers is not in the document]
joelverhagen: Done.

edwin-huber (approved):

joelverhagen:
Are there any other things I should address before the merge? Happy to fix whatever 😃

PR #1393, opened 2022-03-15
Handle new SAS key requiring encryptionScope to be included

### LINKED ISSUE #1389: SAS: Generation for String to Sign does not support versions newer than 2020-10-02
We are trying to update to the latest version of com.azure:azure-storage-blob:12.15.0 and do run into an authentication issue with Azurite. The authentication issue comes down to the part for generating the "string to sign": https://github.com/Azure/Azurite/blob/3ff272f08310f7351ef852c0c12c50b32995c316/src/blob/authentication/IBlobSASSignatureValues.ts#L173

Here all versions past 2020-10-02 are not supported. For reference here is the code com.azure:azure-storage-blob uses to generate this string: https://github.com/Azure/azure-sdk-for-java/blob/f254a1b287d041481df349ae86b362cd9891adfe/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BlobSasImplUtil.java#L304

My request would be whether you could introduce a generateBlobSASSignature-method for versions past 2020-10-02?

MatthewSteeples:
Fixes #1389

Haven't added tests for this yet (the existing code didn't have tests and I'm not yet familiar enough with Typescript projects).

If you'd prefer it was done in a way other than copy and paste then let me know.


MatthewSteeples:
Not sure what's causing that test failure. Doesn't look to be related to my change

blueww:
1. Please also update https://github.com/Azure/Azurite/blob/main/ChangeLog.md.
2. have you tested the changed code works on your side? Normally we should add test case for a code change. 
3. The [account sas ](https://github.com/Azure/Azurite/blob/4290fe9345e64e004ac04f377d2681afca4ea22d/src/blob/authentication/IAccountSASSignatureValues.ts#L152)also need update for the new field: see: https://docs.microsoft.com/en-us/rest/api/storageservices/create-account-sas?redirectedfrom=MSDN#constructing-the-signature-string

MatthewSteeples:
1) Done
2) Done
3) Done

I've also added some tests for the generation of the SAS in the blob utility.

Just out of interest, is there a reason why these methods are re-implemented here as opposed to referencing [the Azure JS SDK](https://github.com/Azure/azure-sdk-for-js)?

blueww:
@MatthewSteeples 
There are some test case filed looks related with your change, see https://dev.azure.com/azure/Azurite/_build/results?buildId=38612&view=logs&jobId=6fc9174b-4a51-5ef4-492c-fd5bfd7cdcf9&j=4af0b868-8cf1-5837-42da-a044ae6eb776&t=073acd1c-6147-5b78-f6f4-911c2bd46705

Like this one:

> 1) Shared Access Signature (SAS) authentication
>        generateAccountSASQueryParameters should work @loki @sql:
>      RestError: <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
> <Error>
>   <Code>AuthorizationFailure</Code>
>   <Message>Server failed to authenticate the request. Make sure the value of the Authorization header is formed correctly including the signature.
> RequestId:57149955-b08c-439b-bb85-8039b19116d6
> Time:2022-03-16T17:18:19.906Z</Message>
> </Error>
>       at new RestError (node_modules/@azure/core-http/dist/index.js:2247:28)
>       at /home/vsts/work/1/s/node_modules/@azure/core-http/dist/index.js:3045:25
>       at runMicrotasks (<anonymous>)
>       at processTicksAndRejections (internal/process/task_queues.js:97:5)

MatthewSteeples:
OK, this is still a work in progress. It looks like the Account SAS method is going to need multiple versions, same as the Blob one does (if statement based on version number passed in).

blueww:
The change of the PR has already been merged with PR https://github.com/Azure/Azurite/pull/1429, with some fix.
The account permission still need to define in Azurite code, as we don't want to introduce more dependency to Azurite dev code.

MatthewSteeples:
Thanks for letting me know

PR #1405, opened 2022-03-25
Additional checks on table creation

### LINKED ISSUE #1355: Emulator doesn't throw error when creating tables with same name but different casing
table

3.15.0

npm

16.13.1

The emulator does not respond with an error when creating a table using an existing name with different casing.

For example, if a table named "table01" exists, I expect an error to be returned by the emulator if I try to create a table named "TABLE01".

According to the [Azure documentation](https://docs.microsoft.com/en-us/rest/api/storageservices/understanding-the-table-service-data-model#table-names), table names are supposed to be case-insensitive.

1. Install and run Azurite.
2. Use the Azure SDK for JavaScript to create a table named "table01":
   ```js
   const client1 = TableClient.fromConnectionString("<DEV CONNECTION STRING>", "table01", { allowInsecureConnection: true });
   await client1.createTable();
   ```
3. Use the Azure SDK for JavaScript to create a table named "TABLE01":
   ```js
   const client2 = TableClient.fromConnectionString("<DEV CONNECTION STRING>", "TABLE01", { allowInsecureConnection: true });
   await client2.createTable();
   ```

No

### LINKED ISSUE #1379: No error when creating a table named "tables"
### Which service (blob, file, queue, table) does this issue concern?
Table

3.14.3

npm

16.13.1

Azurite does not return an error when attempting to create a table named "tables".

### Steps to reproduce the issue

1. Start Azurite
2. Using @azure/data-tables, create a client object connecting to the local emulator.
3. Use the client to create a table named "tables".

No. Azure proper returns an error response. The [documentation](https://docs.microsoft.com/en-us/rest/api/storageservices/understanding-the-table-service-data-model#table-names) also clearly states that "tables" is a reserved table name.

edwin-huber:
Adds additional tests and checks for table names on creation.
Fixes #1379  
Fixes #1355


src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -124,9 +124,11 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
   public async createTable(context: Context, tableModel: Table): Promise<void> {
     // Check for table entry in the table registry collection
     const coll = this.db.getCollection(this.TABLES_COLLECTION);
+    // Azure Storage Service is case insensitive
+    const tableLower = tableModel.table.toLocaleLowerCase();
```
XiaoningLiu, line 128: Somewhere you are using `toLowerCase()` and here use `toLocaleLowerCase()`. Is the difference by design?

src/table/handlers/TableHandler.ts
```diff
@@ -53,7 +53,9 @@ export default class TableHandler extends BaseHandler implements ITableHandler {
     const tableContext = new TableStorageContext(context);
     const accept = this.getAndCheckPayloadFormat(tableContext);
     const account = this.getAndCheckAccountName(tableContext);
-    const table = tableProperties.tableName; // Table name is in request body instead of URL
+    // Table name is in request body instead of URL
+    // we must also force casing as service is case insensitive
+    const table = tableProperties.tableName?.toLowerCase();
```
XiaoningLiu, line 58: If toLowers happen in metadataStore, seems we don't need handler layer do another toLower

src/table/persistence/LokiTableMetadataStore.ts
edwin-huber: No, this was an oversight, changed to toLowerCase()

src/table/handlers/TableHandler.ts
edwin-huber: Moved lower casing on create into MetaDatastore

src/table/persistence/LokiTableMetadataStore.ts
```diff
@@ -160,17 +162,22 @@ export default class LokiTableMetadataStore implements ITableMetadataStore {
   ): Promise<void> {
     // remove table reference from collection registry
     const coll = this.db.getCollection(this.TABLES_COLLECTION);
+    // Azure Storage Service is case insensitive
+    const tableLower = table.toLocaleLowerCase();
     const doc = coll.findOne({
```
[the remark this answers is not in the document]
edwin-huber: Updated to use a regex query to get the table collection, as this can be made case insensitive.
Added additional test to validate logic for delete being case insensitive.

edwin-huber:
I shall change this PR to preserve casing on Tables as per Service behavior.

XiaoningLiu:
> I shall change this PR to preserve casing on Tables as per Service behavior.

Service preserve the Table name casing, and comparison without case, for example:
- Create table "MyTable"
- List Tables should return "MyTable"
- Query/Create/Delete/Update/Merge Entities under table "MyTable" and "mytable" both should work

edwin-huber:
Made changes and added tests to confirm preservation of case on table names.

XiaoningLiu (approved):

PR #1410, opened 2022-03-30
Add more granularity to table etags

### LINKED ISSUE #1286: ETag doesn't change when entity updated in Table service
table

3.15.0

npm

unknown

ETag doesn't change when entity updated in Table service. ETag in table service use timestamp by default. And seems smallest unit of timestamp in table service is 0.001s. If 2 write operations executed fast in a small period of time. ETag may not update properly.

```
await _table.InsertAsync(new DemoEntity("partionKey", "rowKey"), Token);
var e1 = await _table.RetrieveAsync("partionKey", "rowKey", Token);
var e2 = await _table.RetrieveAsync("partionKey", "rowKey", Token);
e2.XXX = "xxx";  // update some value in entity
var result = await _table.MergeAsync(e2, Token);
Assert.AreEqual((int)HttpStatusCode.NoContent, result.HttpStatusCode);  // will pass
var e3 = await _table.RetrieveAsync("partionKey", "rowKey", Token);
Assert.AreEqual("xxx", e3.XXX);  // will pass
Assert.AreNotEqual(e1.ETag, e3.ETag);  // will fail
```
Error message:
```
[INFO] [93]   Error Message:
[INFO] [94]      Expected: not equal to "W/"datetime'2022-01-17T05%3A27%3A10.6390000Z'""
[INFO] [95]   But was:  "W/"datetime'2022-01-17T05%3A27%3A10.6390000Z'""
```

Actually, ETag value will be like `"ETag":"W/\"datetime'2022-01-17T08%3A29%3A47.5220000Z'\"`, part of ETag value is timestamp. However, the accuracy of the timestamp is not enough, is only millisecond level.

This case may pass occasionally, but most time it will fail.
If I sleep some time, for example 1 second, then this case can always pass.

edwin-huber:
Adds ```process.hrtime()``` measurement to etag string for more granularity.
Adds test checking etag validity.
Fixes #1286


src/common/utils/utils.ts
```diff
@@ -89,8 +89,11 @@ export function truncatedISO8061Date(
   // Date.toISOString() will return like "2018-10-29T06:34:36.139Z"
   const dateString = date.toISOString();
 
+  // some clients are very fast, and require more than ms precision available in JS
   return withMilliseconds
-    ? dateString.substring(0, dateString.length - 1) + "0000" + "Z"
+    ? dateString.substring(0, dateString.length - 1) +
+        process.hrtime()[1].toString().padEnd(4, "0").slice(0, 3) +
```
XiaoningLiu, line 95: slice(0, 4) to get 4 digits.

Checked process.hrtime() document.
The absolute value of hrtime() is relative to an arbitrary time in the past.  The string combination between dateString and hrtime() is not a precision calculation of current time in milliseonds.  We should document that clearly.
"hrtime()"

edwin-huber: done

XiaoningLiu (approved):

PR #1540, opened 2022-06-13
info about AZURITE_ACCOUNTS keys `base64(key)`

kinjelom:
There was no information that the key must be base64 encoded in the AZURITE_ACCOUNTS variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -422,18 +422,18 @@ When starting Azurite from npm command line `azurite` or docker image, following
 
 ### Customized Storage Accounts & Keys
 
-Azurite V3 allows customizing storage account names and keys by providing environment variable `AZURITE_ACCOUNTS` with format `account1:key1[:key2];account2:key1[:key2];...`.
+Azurite V3 allows customizing storage account names and keys by providing environment variable `AZURITE_ACCOUNTS` with format `account1:base64(key1a)[:base64(key1b)];account2:base64(key2a)[:base64(key2b)];...`.
 
 For example, customize one storage account which has only one key:
 
 ```cmd
-set AZURITE_ACCOUNTS="account1:key1"
+set AZURITE_ACCOUNTS="account1:a2V5MWE=" # base64(key1a)="a2V5MWE="
 ```
 
 Or customize multi storage accounts and each has 2 keys:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -422,18 +422,18 @@ When starting Azurite from npm command line `azurite` or docker image, following
 
 ### Customized Storage Accounts & Keys
 
-Azurite V3 allows customizing storage account names and keys by providing environment variable `AZURITE_ACCOUNTS` with format `account1:key1[:key2];account2:key1[:key2];...`.
+Azurite V3 allows customizing storage account names and keys by providing environment variable `AZURITE_ACCOUNTS` with format `account1:base64(key1a)[:base64(key1b)];account2:base64(key2a)[:base64(key2b)];...`.
```
blueww, line 425: `account1:base64(key1a)[:base64(key1b)]`, should it be `account1:base64(key1a)[:base64(key2a)]`?

Normally a storage account will have 2 keys: key1 and key2. So we might should also keep the name aligned for key1/key2 of each storage account.

Same as below

kinjelom: If there are two accounts names like `key1` and `key2` suggest the same keys for `account1` and `account2` :)
...but it is of little importance, it is important to show that the keys must be base64 encoded.

blueww:
@[kinjelom](https://github.com/kinjelom)
The key1 & key2 in the doc matches the key1/key2 of a storage account.
Key1/key2 should be base64 encoded string, this is aligned with product Azure.
So if user just use a valid key format in product azure on Azurite, it should work.

Besides that, user need to input based64 encoded key in using client tools to access both Azurite and product azure. 
So the key input into Azurite config is just the key user will use to access Azurite, they are aligned.
After the doc change, I am afraid user will misunderstand the usage, and think Azurite need the key be based64 encode, but access Azurite use plain text key.
So we might should just tell the key should be base64 string, but not use "base64(key1b)" which might misleading user "key1b" is the really key and use to it in azure client tool to access azurite.

blueww:
Close this PR as we already add the info that key should be base64 string with PR https://github.com/Azure/Azurite/pull/1570

PR #1542, opened 2022-06-14
Fix issue where creating a snapshot of a blob fails PutBlock operations which specify a lease.

ljcollins25:
The issue is that snapshot is not specified as empty when finding a blob/block in the Loki blob collection so snapshot instance may be found.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,13 @@
 
 ## Upcoming Release
 
+## 2024.04 Version 3.30.0
+
+General:
+
+- Bump up service API version to 2024-05-04
+- Fixed issue of failure when connecting to mssql with docker image or exe.
+
 Blob:
 
 - Fixed issue of setting blob tag should not update Blob Etag and LastModified. (issue #2327)
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 | Version                                                            | Azure Storage API Version | Service Support                | Description                                       | Reference Links                                                                                                                                                                                                         |
 | ------------------------------------------------------------------ | ------------------------- | ------------------------------ | ------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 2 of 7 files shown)


blueww:
@ljcollins25 

Thanks for raising this PR!
Would you please also update change log and test cases for this change?
Besides that, the PR check has many failures like:
```
src/blob/persistence/LokiBlobMetadataStore.ts(2260,9): error TS1005: ',' expected.
src/blob/persistence/LokiBlobMetadataStore.ts(2277,7): error TS1005: ',' expected.
```

@XiaoningLiu 
Would you please help to look?

ljcollins25 (commented):
Fix mismerge.

ljcollins25:
Unfortunately I'm not a team member so I don't have a good setup to properly land this PR. I'd like to ask a team member to help and do what's necessary to clean up and merge this PR. The changes to the Loki store seem sound but I think I'm missing something with the SQL store.

Thanks!

blueww:
@ljcollins25 

The PR check still fail since test case fail.
As I see, following 2 test cases failed after you change, would you please help to fix them?

The 2 test cases are both in file https://github.com/Azure/Azurite/blob/main/tests/blob/apis/blockblob.test.ts
```
1) BlockBlobAPIs
       stageBlock with double commit block should work @loki @sql:

      AssertionError [ERR_ASSERTION]: 2 == 1
      + expected - actual

      -2
      +1
      
      at Context.it (tests/blob/apis/blockblob.test.ts:188:12)
      at process._tickCallback (internal/process/next_tick.js:68:7)

  2) BlockBlobAPIs
       getBlockList_BlockListingFilter @loki @sql:

      AssertionError [ERR_ASSERTION]: Input A expected to strictly deep-equal input B:
+ expected - actual

- 404
+ 400
      + expected - actual

      -404
      +400
      
      at Context.it (tests/blob/apis/blockblob.test.ts:431:12)
      at process._tickCallback (internal/process/next_tick.js:68:7)

```

Issue #1554, opened 2022-06-19
[Tables] the continuation token on tables is not working when no row key in continuation


bhugot:
### Which service(blob, file, queue, table) does this issue concern?
table
### Which version of the Azurite was used?
latest
### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
github
### What's the Node.js version?
v17.6.0
### What problem was encountered?
When loading entities with continuation token the same result and continuation token is return than the first load if no rowkey continuation token
### Steps to reproduce the issue?

* Insert 21 partitionkey in table
* retrieve the 20 first with a page size of 20 then use the continuation token.
* You are still retrieving the 20 first.
### Have you found a mitigation/solution?
Yes change the code as following:
```
.where((data: any) => {
        if (decodedNextRowKey !== undefined) {
          if (
            data.RowKey >= decodedNextRowKey &&
            (data.PartitionKey === decodedNextPartitionKey ||
              data.PartitionKey === undefined)
          ) {
            return true;
          }
          return false;
        }
        if (decodedNextPartitionKey !== undefined) {
          if (data.PartitionKey >= decodedNextPartitionKey) {
            return true;
          }
          return false;
        }

        return true;
      })
```

PR #1555, opened 2022-06-19
Fixed load multiple page with rows without rowkey

bhugot:
This fix an issue when loading multiple pages of rows that does have empty string rowkey.  #1554

### DIFF
--- a/tests/table/apis/table.entity.issues.test.ts
+++ b/tests/table/apis/table.entity.issues.test.ts
@@ -28,7 +28,7 @@ configLogger(false);
 // Azure Storage Connection String (using SAS or Key).
 const testLocalAzuriteInstance = true;
 
-describe("table Entity APIs test", () => {
+describe("table Entity APIs test : Issues", () => {
   let server: TableServer;
   const tableName: string = getUniqueName("datatables");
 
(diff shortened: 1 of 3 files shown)


tests/table/apis/table.entity.issues.test.ts
```diff
@@ -151,7 +157,7 @@ describe("table Entity APIs test", () => {
       assert.notStrictEqual(result.etag, undefined);
     }
 
-    const maxPageSize = 250; // this should work with a page size of 1000, but fails during response serialization on my machine
+    const maxPageSize = 50; // this should work with a page size of 1000, but fails during response serialization on my machine
```
edwin-huber, line 160: Hi,
could you explain the change to the page size here?  
Was it causing some problem?
Thanks.

edwin-huber (approved):
LGTM - Just wondering if we have an issue with paging performance / serialization... see comment.
@XiaoningLiu

tests/table/apis/table.entity.issues.test.ts
bhugot: Hello, It's just to match the test description 
```should return 101 entities from a paged query at 50 entities per page and single partition```

edwin-huber: Thanks!

XiaoningLiu (approved):

Issue #1565, opened 2022-06-24
Upsert merge and replace return 404 incorrectly when the entity does not exist


joelverhagen:
### Which service(blob, file, queue, table) does this issue concern?

table

### Which version of the Azurite was used?

3.18.0

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)

VS Code extension

### What's the Node.js version?

v16.13.0

### What problem was encountered?

These sort of upsert operations returned 404 when the entity does not yet exist. This is wrong for upsert.

This problem did not occur in 3.17.1.

```csharp
await table.UpsertEntityAsync(entityA, TableUpdateMode.Merge);
await table.UpsertEntityAsync(entityB, TableUpdateMode.Replace);
```

### Steps to reproduce the issue?

I made this helper app to test a bunch of different variants of delete, update, and upsert.

<details>

```csharp
using System.Text.RegularExpressions;
using Azure;
using Azure.Core.Pipeline;
using Azure.Data.Tables;

var client = new TableServiceClient("UseDevelopmentStorage=true", new TableClientOptions
{
    Transport = new HttpClientTransport(new HttpClient(new LoggingHandler { InnerHandler = new HttpClientHandler() }))
});
var table = client.GetTableClient($"test{DateTimeOffset.UtcNow.Ticks}");
table.CreateIfNotExists();

var tests = new Func<int, Task>[]
{
    async i =>
    {
        Console.WriteLine("[New entity, Upsert Merge]");
        var entity = new TableEntity("pk", $"rk{i}");
        await table.UpsertEntityAsync(entity, TableUpdateMode.Merge);
    },
    async i =>
    {
        Console.WriteLine("[New entity, Update Merge *]");
        var entity = new TableEntity("pk", $"rk{i}");
        await table.UpdateEntityAsync(entity, ETag.All, TableUpdateMode.Merge);
    },
    async i =>
    {
        Console.WriteLine("[New entity, Upsert Replace]");
        var entity = new TableEntity("pk", $"rk{i}");
        await table.UpsertEntityAsync(entity, TableUpdateMode.Replace);
    },
    async i =>
    {
        Console.WriteLine("[New entity, Update Replace *]");
        var entity = new TableEntity("pk", $"rk{i}");
        await table.UpdateEntityAsync(entity, ETag.All, TableUpdateMode.Replace);
    },
    async i =>
    {
        Console.WriteLine("[New entity, Delete *]");
[...]

Issue #1493, opened 2022-05-12
Table Service: Replace operation for non-existent entity with wildcard ETag returning 204 instead of 404


ReubenBond:
### Which service(blob, file, queue, table) does this issue concern?

Table

### Which version of the Azurite was used?

Azurite-Table 3.17.1, according to the response headers.

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)

Microsoft Container Registry (mcr.microsoft.com/azure-storage/azurite:latest)

### What's the Node.js version?

Unsure, it's whatever is in that container.

### What problem was encountered?

A discrepancy between Azure Table Storage and Azurite: when using the Azure SDK for .NET and attempting to replace or merge an entity which was never inserted into the table, like so:

``` C#
// NOTE: Merge yields the same erroneous behavior as Replace
var opResult = await Table.UpdateEntityAsync(entity, new ETag("*"), TableUpdateMode.Replace);
```

We are observing that Azurite returns a 204 (inserting the value), but Azure Table Service fails with a 404 as expected (the data is being replaced but is not present to begin with).

### Steps to reproduce the issue?

Something like this, perhaps. If there is a way to enable debug logs for the container, I can do that and report back with them.

```
POST http://127.0.0.1:10002/devstoreaccount1/UnitTestAzureData(PartitionKey='PK-AzureTableDataManagerTests-0655f5a6-0b98-411a-9f87-1904e68719f2',RowKey='RK-003bfce3-5032-4ece-a77a-15822871c500')?$format=application%2Fjson%3Bodata%3Dminimalmetadata HTTP/1.1
x-ms-version: 2019-02-02
DataServiceVersion: 3.0
If-Match: *
Accept: application/json
Content-Type: application/json

{"Data":null,"StringData":"JustData","PartitionKey":"PK-AzureTableDataManagerTests-0655f5a6-0b98-411a-9f87-1904e68719f2","RowKey":"RK-003bfce3-5032-4ece-a77a-15822871c500"}
```

We have a test case which fails on Azurite but not Azure Table Service here: https://github.com/dotnet/orleans/blob/4570a2a5a8a7ca65b13031010fc994bcfede484f/test/Extensions/TesterAzureUtils/AzureTableDataManagerTests.cs#L77

### Have you found a mitigation/solution?

We can disable these tests when Azurite is being used.

edwin-huber:
Hi,
are you sure that this is the root cause of the error?
[...]

PR #1566, opened 2022-06-24
Fix upsert merge to not throw a 404 when the entity is missing

joelverhagen:
If-Match: * and no If-Match header behaves differently. If-Match: * requires the entity to exist.

Address https://github.com/Azure/Azurite/issues/1565 as well as another pre-existing problem where Batch Update acted like an Upsert.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,10 @@
 
 ## Upcoming Release
 
+Table:
+- Correctly responds with status 202 on merge with non-existent entity.
+- Properly differentiate between upsert and update in batch merge and replace.
+
 ## 2022.06 Version 3.18.0
 
 General:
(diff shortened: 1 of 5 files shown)


joelverhagen:
@edwin-huber, my PR seems somewhat contradictory to your PR https://github.com/Azure/Azurite/issues/1493. I am a bit confused about the whole situation but my change seems right given the test app I included in my bug report: https://github.com/Azure/Azurite/issues/1565.

edwin-huber:
Sorry for the confusion.

Unfortunately there are differences in behaviors between the various SDKs, not just language SDKs, but also SDKs for JavaScript.

We need to review the test suite and make sure that we have correct coverage and validate against the service.
Worst case we need to write RAW Http requests to test the API behavior if the behavior is not reproducible due to TypeScript / JavaScript SDK limitations.

I'll write up an issue as specs to address this properly, will link the bugs and get clarity on the behavior, and will base the work on this PR.

edwin-huber:
I should add that we have been toying with the idea of including tests written in other languages in the Azure Pipelines, which would make ensuring proper coverage easier.

joelverhagen:
I'd be happy to contribute some integrations with the Azure SDK for .NET. It's not immediately clear to me where to put mixed coding languages in a single repo... maybe just under the `tests` repo? If this is better left to you and the team that's okay too. If there's a starting point I can add to it as well. 

All in a separate PR of course :)

edwin-huber (approved):
Changes look good.
Put through additional tests, which will be added via a separate PR.

Issue #1579, opened 2022-07-02
Missing etag in header for GET Operation


yshahin:
### Which service(blob, file, queue, table) does this issue concern?

table

### Which version of the Azurite was used?

3.18.0

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)

npm, DockerHub

### What's the Node.js version?

18.4.0

### What problem was encountered?

GET Entity with PartitionKey= and RowKey= does not return etag in header

### Steps to reproduce the issue?

Create entry in azure table with any partition_key and row_key, then perform an http GET request against the entry created
```
http://127.0.0.1:10002/devstoreaccount1/aztest(PartitionKey='partition_key',RowKey='row_key')
```
and check the response headers. etag header is not returned

### Have you found a mitigation/solution?

Solution is to modify TableHandler.ts line 686
```
    const response: Models.TableQueryEntitiesWithPartitionAndRowKeyResponse = {
      statusCode: 200,
      date: tableContext.startTime,
      clientRequestId: options.requestId,
      requestId: context.contextID,
      version: TABLE_API_VERSION
    };
```

to

```
    const response: Models.TableQueryEntitiesWithPartitionAndRowKeyResponse = {
      statusCode: 200,
      date: tableContext.startTime,
      clientRequestId: options.requestId,
      requestId: context.contextID,
      eTag: entity.eTag,
      version: TABLE_API_VERSION
    };
```

edwin-huber:
Many thanks for raising this.
Someone is working on it.

PR #1613, opened 2022-07-25
Fix issue 1579: Missing etag in header for GET Operation

maxisl:
Fix issue [#1579](https://github.com/Azure/Azurite/issues/1579)

Added test case to check specifically for etag in headers in table.entity.rest.test.ts

and modified TableHandler.ts in line 686 to solve issue


tests/table/apis/table.entity.azure.data-tables.test.ts
```diff
@@ -1698,4 +1698,61 @@ describe("table Entity APIs test - using Azure/data-tables", () => {
 
     await tableClient.deleteTable();
   });
+
+  // pseudo
```
edwin-huber, line 1702: Can you please remove the pseudo code comments?
thanks.

edwin-huber: It also looks like these are unrelated to the issue you are fixing.

edwin-huber (changes requested):
Please remove Doc changes from this PR and the changes to tests/table/apis/table.entity.azure.data-tables.test.ts (as I don't think they are intended for this PR)

PR #1755, opened 2022-11-17
Resolve tests timeout and closes connections

giggio:
This is related to https://github.com/nodejs/node/issues/2642
Added stoppable module and stops the server.
Also added --exit to the tables test, so it does not hang.
Both conditions were holding the test process indefinitely.


XiaoningLiu (approved):

XiaoningLiu:
@blueww @EmmaZhu  We need regenerate license for new dependency.

PR #1801, opened 2023-02-09
Support batch request formatting from Azure-Storage/9.3.2

### LINKED ISSUE #1798: CloudTable.ExecuteBatch failing with 500 error
Table

3.19.0

Visual Studio Enterprise 2022

16.16.0

We are trying to migrate from Azure Storage Emulator to Azurite and are noticing lots of UTs are failing with 500 error when CloudTable.ExecuteBatch is called using Azurite. 

Start Azurite and execute code calling CloudTable.ExecuteBatch
The error returned is 500 URI Error: URI malformed

If possible, please provide the debug log using the -d parameter, replacing \<pathtodebuglog\> with an appropriate path for your OS, or review the instructions for docker containers:

```
-d "<pathtodebuglog>"
```

[debug5.log](https://github.com/Azure/Azurite/files/10483307/debug5.log)

Please be sure to remove any PII or sensitive information before sharing!  
The debug log will log raw request headers and bodies, so that we can replay these against Azurite using REST and create tests to validate resolution.

No

edwin-huber:
fixes #1798

Includes test to validate fix.

Adds safer URI and URI component decoding to the batch orchestrator when processing some batch requests from Azure-Storage package.


blueww:
@edwin-huber 
Thanks for raising the fix PR! 
The PR generally looks good to me.

Just a small concern:  If the batch sub request encoded Uri contains "%2525" (e.g.: original partition key contains "%25"), will this change cause the Uri be decoded 2 times and get incorrect "%", instead of original "%25", then cause the sub request fail since the partition key not exist.
Do you think this is a valid scenario?

src/table/batch/TableBatchSerialization.ts
```diff
@@ -538,10 +538,14 @@ export class TableBatchSerialization extends BatchSerialization {
     }
     const trimmedUrl: string = request.getUrl().substring(0, offsetPosition);
     let entityPath = trimmedUrl + "(PartitionKey='";
-    entityPath += request.params.tableEntityProperties!.PartitionKey;
+    entityPath += encodeURIComponent(
```
EmmaZhu, line 541: Seems the entityPath is not in response URL. Do we need to encode a string outside the URL?  Does storage service encode it in this case?

tests/table/apis/table.entity.azure.data-tables.test.ts
```diff
@@ -1038,4 +1039,245 @@ describe("table Entity APIs test - using Azure/data-tables", () => {
...
+    const tableClient = createAzureDataTablesClient(
+      testLocalAzuriteInstance,
+      getUniqueName("percentBatch")
+    );
+    await tableClient.createTable();
+    const percentPartition = "percentRowBatch";
+    const testEntities: TableTestEntity[] = [
+      // { rowKey: "@r%25o%0123512512356", partitionKey: percentPartition },
+      // { rowKey: "r%25131463146346", partitionKey: percentPartition },
+      // { rowKey: "@r%25o%21235136314613", partitionKey: percentPartition }
```
blueww, line 1053: Does it cover scenario contains partitionkey or  rowkey include string can be successfully decode 2 times like "%2525"

edwin-huber: yes, the test ensures that strings are then decoded to %<digit string>
I also specifically ran the test on %2525 and it also passes.

blueww: Thanks @edwin-huber  for the confirmation!
Just curious which case can cover the %2525 testing? 
I don't fine such a string in the test case? 

To avoid the Uri be duplicated decoded, it's better we can have such a case.

src/table/batch/TableBatchSerialization.ts
edwin-huber: Yes, the Uri in the metadata should be encoded.
You will see this in the old Azure Storage SDK if you perform operations using OData Full Metadata option.

tests/table/apis/table.entity.azure.data-tables.test.ts
edwin-huber: This is covered by Line 1059
I also ran a local test just now where I confirmed %2525 explicitly, rather than randomly generated strings.
We can force a test on %2525 if you feel it is needed.
```
testEntities[1].rowKey = testEntities[1].rowKey.replace("row", "r%25");
```

blueww: Great! Thanks!

blueww (approved):

EmmaZhu (approved):

PR #1807, opened 2023-02-16
Allow query on table name starting with underscore

### LINKED ISSUE #1806: [BUG] query with property that has and underscore ('_') fails in version >= 3.21.0

Tablee

3.21.0 & 3.22.0

nom

18.13.0

When executing any query with a property that has an underscore (e.g. "_foo eq bar"), an error is returned.

Using Azure Storage Explorer
image

1. Create a table and an entity with an underscore (the entity doesn't matter, error occurs even if you have no entities).
image

2. Query with underscore
image

3. See error
4. Query with NO underscore succeeds
image

Revert to 3.20.1

edwin-huber:
Adds tests and support for underscore to table prop query.
fixes #1806


blueww (approved):
Thanks for the quick fix!
LGTM

PR #1820, opened 2023-02-23
support developing durable functions with a local Azure Functions 

kentis:
Sometimes the local Azure Functions host will set messagettl to −1 instead of -1 (the first numer uses unicode letter U+2212 instead of the regular '-' character). This causes Azurite to, understandiby, return a 400 error and the durable function to not start. By adding the code here this is mitigated by replacing '−' with '-' in messagettl query paramters before they are attemted to be used and thus avoiding retuinijng 400.

### PR Branch Destination

### Always Add Test Cases

QueueServiceClient does not allow adding strings as messagettl. Could you please advice on how to create the test case in this instance? :)

### Add Change Log

### Development Guideline

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,10 @@
 
 ## Upcoming Release
 
+Queue:
+
+- Fixed issue with running Durable Functions from Azure Functions where the Functions would send oddly an formatted negative number the messagettl parameter.
+
 ## 2023.02 Version 3.22.0
 
 General:
--- a/src/queue/QueueRequestListenerFactory.ts
+++ b/src/queue/QueueRequestListenerFactory.ts
@@ -47,7 +47,7 @@ export default class QueueRequestListenerFactory
     private readonly skipApiVersionCheck?: boolean,
     private readonly oauth?: OAuthLevel,
     private readonly disableProductStyleUrl?: boolean
-  ) {}
+  ) { }
 
   public createRequestListener(): RequestListener {
     const app = express().disable("x-powered-by");
@@ -95,6 +95,14 @@ export default class QueueRequestListenerFactory
     // Manually created middleware to deserialize feature related context which swagger doesn't know
     app.use(createQueueStorageContextMiddleware(this.skipApiVersionCheck, this.disableProductStyleUrl));
 
+    // Sometimes Azure Functions sends messagettl as -1 with a unicode minus sign (−)
+    app.use((req, res, next) => {
+      if (req && req.query && req.query.messagettl) {
+        let messagettl = req.query.messagettl?.replace("−", "-");
+        req.query.messagettl = messagettl;
+      }
+      next();
+    })
     // Dispatch incoming HTTP request to specific operation
     app.use(middlewareFactory.createDispatchMiddleware());
 
--- a/src/queue/QueueRequestListenerFactory.ts
+++ b/src/queue/QueueRequestListenerFactory.ts
@@ -47,7 +47,7 @@ export default class QueueRequestListenerFactory
     private readonly skipApiVersionCheck?: boolean,
     private readonly oauth?: OAuthLevel,
     private readonly disableProductStyleUrl?: boolean
-  ) { }
+  ) {}
 
   public createRequestListener(): RequestListener {
     const app = express().disable("x-powered-by");
--- a/src/queue/QueueRequestListenerFactory.ts
+++ b/src/queue/QueueRequestListenerFactory.ts
@@ -98,11 +98,12 @@ export default class QueueRequestListenerFactory
     // Sometimes Azure Functions sends messagettl as -1 with a unicode minus sign (−)
     app.use((req, res, next) => {
       if (req && req.query && req.query.messagettl) {
-        let messagettl = req.query.messagettl?.replace("−", "-");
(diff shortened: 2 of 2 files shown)


kentis:

blueww (approved):
LGTM

blueww:
@kentis 

I just tested on Azure product Storage server, and it looks not work with "−1" but only work with "-1".
Can you pass test with "−1" on product Azure?
If not, we should not take this PR change, but the durable function should be fixed?


- request fail with "−1" on product Azure
```
POST https://[accountName].queue.core.windows.net/weitest/messages?sv=2019-07-07&sig=[hidden]&st=2023-02-25T16:00:00Z&se=2024-03-02T16:00:00Z&sp=rap&api-version=2019-07-07&messagettl=%E2%88%921 HTTP/1.1
User-Agent: Mozilla/5.0 (Windows NT; Windows NT 10.0; zh-CN) WindowsPowerShell/5.1.19041.2364
Content-Type: application/x-www-form-urlencoded
Host: [accountName].queue.core.windows.net
Content-Length: 118
Expect: 100-continue

<?xml version="1.0" encoding="utf-8"?><QueueMessage><MessageText>VGhpcyBpcyBtZXNzYWdlIDE=</MessageText></QueueMessage>
HTTP/1.1 400 Value for one of the query parameters specified in the request URI is invalid.
Content-Length: 413
Content-Type: application/xml
Server: Windows-Azure-Queue/1.0 Microsoft-HTTPAPI/2.0
x-ms-request-id: d05102bd-4003-0047-0e84-4ab199000000
x-ms-version: 2019-07-07
x-ms-error-code: InvalidQueryParameterValue
Date: Mon, 27 Feb 2023 08:21:22 GMT

﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>InvalidQueryParameterValue</Code><Message>Value for one of the query parameters specified in the request URI is invalid.
RequestId:d05102bd-4003-0047-0e84-4ab199000000
Time:2023-02-27T08:21:22.4939252Z</Message><QueryParameterName>messagettl</QueryParameterName><QueryParameterValue>−1</QueryParameterValue><Reason>Must be an integer value.</Reason></Error>
```

- request pass with "-1" on product Azure
```
POST https://[accountName].queue.core.windows.net/weitest/messages?sv=2019-07-07&sig=[hidden]&st=2023-02-25T16%3A00%3A00Z&se=2024-03-02T16%3A00%3A00Z&sp=rap&api-version=2019-07-07&messagettl=-1 HTTP/1.1
User-Agent: Mozilla/5.0 (Windows NT; Windows NT 10.0; zh-CN) WindowsPowerShell/5.1.19041.2364
Content-Type: application/x-www-form-urlencoded
Host: [accountName].queue.core.windows.net
Content-Length: 118
Expect: 100-continue

<?xml version="1.0" encoding="utf-8"?><QueueMessage><MessageText>VGhpcyBpcyBtZXNzYWdlIDE=</MessageText></QueueMessage>
HTTP/1.1 201 Created
Transfer-Encoding: chunked
Content-Type: application/xml
Server: Windows-Azure-Queue/1.0 Microsoft-HTTPAPI/2.0
x-ms-request-id: 183c0fe6-5003-005b-2c84-4ae3f9000000
x-ms-version: 2019-07-07
Date: Mon, 27 Feb 2023 08:20:15 GMT

197
﻿<?xml version="1.0" encoding="utf-8"?><QueueMessagesList><QueueMessage><MessageId>39d8651d-c83c-48a0-9ce1-cc18f148a0d1</MessageId><InsertionTime>Mon, 27 Feb 2023 08:20:16 GMT</InsertionTime><ExpirationTime>Fri, 31 Dec 9999 23:59:59 GMT</ExpirationTime><PopReceipt>AgAAAAMAAAAAAAAAMm4+UoRK2QE=</PopReceipt><TimeNextVisible>Mon, 27 Feb 2023 08:20:16 GMT</TimeNextVisible></QueueMessage></QueueMessagesList>
0
```

blueww (changes requested):
"−1" can‘t pass test on product Azure, so we should not take this change.

kentis:
This means that devs will have to locally patch versions of Azurite, or add code only for local testing, since DurableTask also seems reluctant to fix trhe issue ref: https://github.com/Azure/durabletask/issues/809

This is much more of an issue when running locally than in Azure product since Culture typically varies way more on developer machines than it does in production.

kentis:
> "−1" can‘t pass test on product Azure, so we should not take this change.

@blueww  What if we hid the functionality behind a command line argument? Would that make the change more acceptable?

blueww:
@kentis 

Since product Azure doesn't accept "−1", if Azurite accept it, there might be some customer whose program with "−1" can pass with Azurite, but fail with product Azure. 
And for easy customer usage, we will try to keep the parameter count as low as possible, and only add parameter when it's really necessary.

For this case, it's actually a bug of Azure Function (I don't see it repro on any other SDK, please correct me if not.).  
The correct way to resolve it is: Fix the bug in Azure functions. 
And before the issue is fixed in Azure Function, a workaround is patch Azurite locally, and run with your local built Azurite. (Or patch Azure Function?)

blueww:
@kentis

I will close this PR.
Please fix Azure Function for this issue, as it will also repro on product Azure.

PR #1861, opened 2023-03-24
Fix for user delegation key when uploading a blob from a container SAS

seankennett:
Bug found when you create a container SAS token with write permissions via a user delegated key.  When you upload a blob then it breaks as signatures do not match up.  Behaves fine for things like get blobs as there isn't a blob name added to the string hash.

Saw same fix on generateBlobSASSignature20150405 so followed.


seankennett:

EmmaZhu (approved):

Issue #1854, opened 2023-03-22
Azure Durable Function does not finish when I configured Azurite with multiple Storage Accounts.


ecormaksin:
### Which service(blob, file, queue, table) does this issue concern?

table

### Which version of the Azurite was used?

3.22.0

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)

DockerHub

### What's the Node.js version?

v14.17.0 (in Docker container)

### What problem was encountered?

Firstly, I posted [the same question to Microsoft Q&A](https://learn.microsoft.com/en-us/answers/questions/1189770/azure-durable-function-does-not-finish-when-i-conf). Then, the respondent (your company's  staff) advised me to post an issue here.

I'm developing Azure Function (Python).

And I configuire [Azurite with multiple Storage Accounts & Keys](https://github.com/Azure/Azurite#customized-storage-accounts--keys-1).
I use Azurite from container, so I configure `docker-compose.yml` as follows:

```yaml
# (omission)
services:
  azurite:
    image: mcr.microsoft.com/azure-storage/azurite
    ports:
      - 10000:10000
      - 10001:10001
      - 10002:10002
    environment:
      - AZURITE_ACCOUNTS=local-dev-sa-001:cGFzc3dvcmQ=;local-dev-sa-003:cGFzc3dvcmQ=;
    # (omission)
```

And I configure `local.settings.json` in Azure Function code as follows (other value is omitted):

```json
{
  "Values": {
    "AzureWebJobsStorage": "DefaultEndpointsProtocol=http;AccountName=local-dev-sa-001;AccountKey=cGFzc3dvcmQ=;BlobEndpoint=http://127.0.0.1:10000/local-dev-sa-001;QueueEndpoint=http://127.0.0.1:10001/local-dev-sa-001;TableEndpoint=http://127.0.0.1:10002/local-dev-sa-001;"
  }
}
```

When I launch Azure Function App from Visual Studio Code with `F5` and execute Durable Funcion, the funcion does not finish. Not only that, the function is executed multiple times. I confirmed from the log output on Visual Studio Code terminal pane.

As for Azure Function, errors do not seem to happen.
When I access `statusQueryGetUri` of the Durable Function response, `runtimeStatus` is `Pending`.

The same code works fine when I use Azurite default configuration.

### Steps to reproduce the issue?

I launched Azurite with [debug log option](https://github.com/Azure/Azurite#debug-log-configuration).

Errors seem to happen as follows:

[...]

PR #1862, opened 2023-03-24
Update README.md

ecormaksin:
according to [this GitHub issue](https://github.com/Azure/Azurite/issues/1854)

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,15 +460,14 @@ Azurite will refresh customized account name and key from environment variable e
 
 > Note. Default storage account `devstoreaccount1` will be disabled when providing customized storage accounts.
 
-> Note. When changing storage account name, keep these rules in mind as same as [Azure Storage Account](https://learn.microsoft.com/en-us/azure/storage/common/storage-account-overview#storage-account-name):
->  - Storage account names must be between 3 and 24 characters in length and may contain numbers and lowercase letters only.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -460,6 +460,10 @@ Azurite will refresh customized account name and key from environment variable e
 
 > Note. Default storage account `devstoreaccount1` will be disabled when providing customized storage accounts.
 
+> Note. When changing storage account name, keep these rules in mind as same as [Azure Storage Account](https://learn.microsoft.com/en-us/azure/storage/common/storage-account-overview#storage-account-name):
+>  - Storage account names must be between 3 and 24 characters in length and may contain numbers and lowercase letters only.
+>  - Your storage account name must be unique within Azurite. No two storage accounts can have the same name.
```
blueww, line 465: 1. It's better to move this note as the last note in this section for : 1. It might not a common case for customer to use an invalid storae account name, I only see it once. 2. the note is long.

2. I think we can remove following statement, since the string to define the account name is not so long, customer normally won't use duplicated name. (In product Azure, this is necessary, since storage account names are from different customers.)

> - Your storage account name must be unique within Azurite. No two storage accounts can have the same name.

ecormaksin: Thank you for yor review. I fixed.

blueww (approved):

PR #1892, opened 2023-04-01
add support for for datalake api and implemented some methods in it in both Loki and SQL

mahmoudbahaa:
The PR was manually tested didn't have enough time to add test cases. may add some in the future if I had the time. or if someone want to do that go ahead.


mahmoudbahaa:
i will rebase and then reopen

Issue #553, opened 2020-09-10
Is there a plan to support AdlsGen2 (Datalake Storage) on top of blobstore emulator ? 


Arnaud-Nauwynck:
### Which service(blob, file, queue, table) does this issue concern?
not existing currently in V3 : AdlsGen (Datalake)

This is a question: Is there a plan to support AdlsGen2 (Datalake Storage) on top of blobstore emulator ? 
If yes, when would it be available?

### Which version of the Azurite was used?
V3, unusable yet, laking ADlsGen2 support

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
npm

### Have you found a mitigation/solution?
no
not able to develop on local with emulator, forced to connect to azure

XiaoningLiu:
Hi @Arnaud-Nauwynck, can you further description your usage scenarios? For example, are you using a datalake gen2 account with or without hierarchical namespace enabled? Do you use any datalake storage sdks? which datalake gen2 features you are most interested? Interop between blob or datalake gen2?

rh99:
As for me, I'd like to Spark and hadoop-azure, as described here: https://stackoverflow.com/questions/65050695/how-can-i-read-write-data-from-azurite-using-spark

yuranos:
Hi @XiaoningLiu 
We are using Gen 2 with Hierarchical namespace enabled.
We write our apps with Akka Streams and need to write to and read from to Datalake Gen 2.
Akka Streams has [HDFS](https://doc.akka.io/docs/alpakka/current/hdfs.html) connector and since Datalake Gen 2 is HDFS-compatible, so far so good.

However, we need to add test coverage to our code. And not just basic unit tests, but something that can verify that integration is Ok end to end, but without being tightly coupled to network for real calls to Azure.

We implemented what we needed with [testcontainers](https://www.testcontainers.org/), but hadoop inside docker is a bit tricky to use and requires quite unconventional setup(with host resolutions, some env vars, permissions, etc.) while all we need is the filesystem for tests.

KlaudiuszBryjaRelativity:
When do you release ADLS feature?
BTW do you have plan to add support for file share?

karimdabbagh:
Any updates on this?

XiaoningLiu:
Hi guys, we get your feedbacks and the asks to support datalake gen2 on Azurite. We do think it's a valid ask and keep it open  for collecting requirements and feedbacks. Unluckily, it's not our current priorities yet.

[...]

PR #1893, opened 2023-04-01
add support for for datalake api

mahmoudbahaa:
The PR was externally tested (through java unit tests that use Hadoop File System and the abfs:// scheme and Azure sdk) 

Didn't have enough time to add test cases to the repo itself. May add some in the future if I had the time. or if someone want to do that go ahead.

Address issue #553


mahmoudbahaa:
@microsoft-github-policy-service agree company="Incorta"

mahmoudbahaa:
fix existing test cases by merging different conflicting apis into one and reset enable xml for all except listpaths (done by manual change in generated code by changing isXML of the opration to false)

blueww:
@mahmoudbahaa 

Thanks so much for raising this PR!
We will look into it and update you later.

However, with a glace review, I see some problems:

1. The PR doesn't introduce a new endpoint dfs, all dfs requests are handle by blob endpoint.
This will take problem: 
- It will cause 2 blob/dfs request which has simialr parameter like [Path_GetProperties](https://learn.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/get-properties) and [Blob_GetProperties](https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties?tabs=azure-ad) will handled by same code, but actually in rest spec they have different parameter and responds.
- This will take regression to exist blob API, so exist Azurite user on blob API might be blocked.

2. And even Blob endpoint API will work differently on HNS enabled account and not enabled account. So to support both account type, we might should add a new start parameter to Azurite to indicate the account enabled HNS or not. (To avoid regression, default should be not enabled.)

mahmoudbahaa:
@blueww 

reagrding your commments:

> The PR doesn't introduce a new endpoint dfs, all dfs requests are handle by blob endpoint.

I introduced new endpoint named dfs on port 10003 which support both blob and dfs requests.

> It will cause 2 blob/dfs request which has simialr parameter like [Path_GetProperties](https://learn.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/get-properties) and [Blob_GetProperties](https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties?tabs=azure-ad) will handled by same code, but actually in rest spec they have different parameter and responds.

I merged both parameters and responds in these cases into the dfs one and removed the blob one

> This will take regression to exist blob API, so exist Azurite user on blob API might be blocked.

Changes made to blob endpoint are now minimal

> And some Blob endpoint API will work differently on HNS enabled account and not enabled account. So to support both account type, we might should add a new start parameter to Azurite to indicate the account enabled HNS or not. (To avoid regression, default should be not enabled.)

for now the HNS is enabled by default in dfs endpoint since some endpoints don't work for non-HNS enabled and in datalake gen2 it makes sense to make HNS enabled by default.

> Please write doc for which features/API has already supported, and which has not, this is very important for customer to using the feature in Azurite, and for the following improvement/trouble shooting. You can refer https://github.com/Azure/Azurite#support-matrix for how to write it.

done

> Path_SetAccessControl: not implemented

done but the recursive one is not supported yet as in the support matrix

> Datalake SAS not work (include "sr=d","sdd=")

should be working now

> Path_Create : header "x-ms-permissions" and "x-ms-umask" works not well (with "x-ms-permissions: rwxrwxrwx" and "x-ms-umask: 0070", Public Azure will get "rwx---rwx", but current Azurite get "rwxrwxr--")

should be working properly now

> FileSystem_ListPaths/Path_GetProperties:
> list from filesystem root failed with 404
> Result : missing many important properties like permission/group, 

permission/group is now included in both

> FileSystem_ListPaths has additional "isDirectory":false which will cause .net SDK fail,

actually according to spec in listPaths each Path should have have isDirectory flag https://learn.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/list#path

> We don't allow manually update auto-generated code, or generate code again with new swagger will overwrite the manual change. The correct way is to update swagger and record the changes in Readme.md (better with directive), or update the generate code tool.

I did almost all needed changes to swagger and documented in dfs.md file that generate the swagger. only remaining changes needed that is not included in the swagger but are also documented in dfs.md:
1) isXML is set to false in listPaths (generator should respect accept property of the request and serialize according to it) [needed for hadoop]
2) add req.getHeader("X-HTTP-Method-Override") in dispatch.middleware.ts since it can be used instead of req.getHeader("X-HTTP-Method") sometimes [also needed for hadoop]

as far as i can tell the actual generator code is private so don't have access to it to change those

> As you have mentioned, test case still not added. But enough test coverage is necessary to merge the PR, to ensure quanlity and avoid regression.

test cases was added specifically for datalake under tests/dfs/storage-file-datalake

> Missing some necessary doc changes like Readme, change log ...

added needed changes for readme

please check if other changes is needed

blueww:
@mahmoudbahaa 

Thanks for the update!

As this is a big PR we might need more time to review it and will reply you later.

For the `"isDirectory":false`, when list path from server it won't return for file. See following responds from product server in my test.
And one important thing is return it for file  will make[ .net sdk ](https://www.nuget.org/packages/Azure.Storage.Files.DataLake/12.14.0-beta.1)fail with error "The requested operation requires an element of type 'String', but the target element has type 'False'."
```
{"paths":[
{"contentLength":"0","creationTime":"133256687100169757","etag":"0x8DB3A575D040E1D","expiryTime":"0","group":"$superuser","isDirectory":"true","lastModified":"Tue, 11 Apr 2023 06:38:30 GMT","name":"dir1","owner":"$superuser","permissions":"rwx---rwx"},
{"contentLength":"1024","creationTime":"133256687147652690","etag":"0x8DB3A5760706833","expiryTime":"0","group":"$superuser","lastModified":"Tue, 11 Apr 2023 06:38:35 GMT","name":"dir1/__datetime=2020-09-14 09%3A00%3A00","owner":"$superuser","permissions":"rwx---rwx"},
{"contentLength":"1024","creationTime":"133256687195188636","etag":"0x8DB3A5762ADED9C","expiryTime":"0","group":"$superuser","lastModified":"Tue, 11 Apr 2023 06:38:39 GMT","name":"dir1/f%3Aile2","owner":"$superuser","permissions":"rw-r-----"},
{"contentLength":"1024","creationTime":"133256687207598063","etag":"0x8DB3A57636B47EF","expiryTime":"0","group":"$superuser","lastModified":"Tue, 11 Apr 2023 06:38:40 GMT","name":"dir1/testfile_1K_0","owner":"$superuser","permissions":"rw-r-----"},{"contentLength":"0","creationTime":"133256687133364901","etag":"0x8DB3A575EFE92A5","expiryTime":"0","group":"$superuser","isDirectory":"true","lastModified":"Tue, 11 Apr 2023 06:38:33 GMT","name":"dir2","owner":"$superuser","permissions":"rwxr-x---"}]}
```

mahmoudbahaa:
@blueww 

- rebased the PR 
- squashed the commits
- addressed the isDirectory issue by setting it to undefined if it is false (since the default is false as per the specs)

mahmoudbahaa:
@blueww I restructed the PR to reduce duplicate code between blob and dfs end points and splitted it into 2 phases

* phase 1 https://github.com/Azure/Azurite/pull/1933
* phase 2 https://github.com/Azure/Azurite/pull/1934

Issue #1872, opened 2023-03-28
[Task] [Blob Tag] Phase I - Implement BlobTag on API: GetBlob, GetBlobProperties, PUTBlob, ListBlob, PutBlockList, CopyBlob


blueww:
Return Blob tag in: GetBlob, GetBlobProperties, ListBlob (when related query parameter is specified)
Set Blob Tag in: PUTBlob, PutBlockList, CopyBlob

blueww:
The code is PR : https://github.com/Azure/Azurite/pull/1912

blueww:
already released

Issue #1871, opened 2023-03-28
[Task] Implement API: GetBlobTag, SetBlobTag


blueww:
Implement Following API and the related database change: 
https://github.com/Azure/Azurite/blob/a1527fc0ec75137de6982b8dd3ce60fd38f9504d/src/blob/handlers/BlobHandler.ts#L1234
https://github.com/Azure/Azurite/blob/a1527fc0ec75137de6982b8dd3ce60fd38f9504d/src/blob/handlers/BlobHandler.ts#L1241

blueww:
The API implemetation draft is in: https://github.com/blueww/Azurite/tree/blobtag1

Just do sanity check, still miss SAS, test case, change log.

blueww:
Already released

PR #1912, opened 2023-04-12
[Blob Tag] Phase I - Implement SetTag, GetTag on differenet blob APIs

blueww:
This is a PR for Blob tag Phase I change:

Implement set/get BlobTag on API: GetBlobTag, SetBlobTag, GetBlob, GetBlobProperties, PUTBlob (Block/page/append), ListBlob, PutBlockList, copyFromURL, startCopyFromURL.

https://github.com/Azure/Azurite/issues/1871
https://github.com/Azure/Azurite/issues/1872

### PR Branch Destination

### Always Add Test Cases

### Add Change Log

### Development Guideline

### DIFF
--- a/src/blob/persistence/LokiBlobMetadataStore.ts
+++ b/src/blob/persistence/LokiBlobMetadataStore.ts
@@ -62,6 +62,7 @@ import IBlobMetadataStore, {
   SetContainerAccessPolicyOptions
 } from "./IBlobMetadataStore";
 import PageWithDelimiter from "./PageWithDelimiter";
+import { getBlobTagsCount, getTagsFromString } from "../utils/utils";
 
 /**
  * This is a metadata source implementation for blob based on loki DB.
@@ -1199,6 +1200,8 @@ export default class LokiBlobMetadataStore
       context
     );
 
+    doc.properties.tagCount = getBlobTagsCount(doc.blobTags);
+
     return {
       properties: doc.properties,
       metadata: doc.metadata,
@@ -1913,7 +1916,8 @@ export default class LokiBlobMetadataStore
       leaseBreakTime:
         destBlob !== undefined ? destBlob.leaseBreakTime : undefined,
       committedBlocksInOrder: sourceBlob.committedBlocksInOrder,
-      persistency: sourceBlob.persistency
+      persistency: sourceBlob.persistency,      
+      blobTags: options.blobTagsString === undefined ? undefined : getTagsFromString(options.blobTagsString, context.contextId!)
     };
 
     if (
@@ -2099,7 +2103,8 @@ export default class LokiBlobMetadataStore
       leaseBreakTime:
         destBlob !== undefined ? destBlob.leaseBreakTime : undefined,
       committedBlocksInOrder: sourceBlob.committedBlocksInOrder,
-      persistency: sourceBlob.persistency
+      persistency: sourceBlob.persistency,      
+      blobTags: options.blobTagsString === undefined ? undefined : getTagsFromString(options.blobTagsString, context.contextId!)
     };
 
     if (
@@ -2506,6 +2511,7 @@ export default class LokiBlobMetadataStore
       doc.properties.contentEncoding = blob.properties.contentEncoding;
       doc.properties.contentLanguage = blob.properties.contentLanguage;
       doc.properties.contentDisposition = blob.properties.contentDisposition;
+      doc.blobTags = blob.blobTags;
       doc.properties.contentLength = selectedBlockList
         .map((block) => block.size)
         .reduce((total, val) => {
--- a/src/blob/persistence/LokiBlobMetadataStore.ts
+++ b/src/blob/persistence/LokiBlobMetadataStore.ts
@@ -3341,8 +3341,6 @@ export default class LokiBlobMetadataStore
       true
     );
 
-    validateWriteConditions(context, modifiedAccessConditions, doc);
-
     if (!doc) {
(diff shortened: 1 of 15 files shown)


src/blob/persistence/LokiBlobMetadataStore.ts
```diff
@@ -3299,6 +3306,103 @@ export default class LokiBlobMetadataStore
...
+      account,
+      container,
+      blob,
+      snapshot,
+      context,
+      false,
+      true
+    );
+
+    validateWriteConditions(context, modifiedAccessConditions, doc);
```
XiaoningLiu, line 3344: Does set tag support access conditions?

src/blob/persistence/SqlBlobMetadataStore.ts
```diff
@@ -289,6 +290,9 @@ export default class SqlBlobMetadataStore implements IBlobMetadataStore {
         },
         metadata: {
           type: "VARCHAR(2047)"
+        },
+        blobTags: {
+          type: "VARCHAR(2047)"
```
XiaoningLiu, line 295: What's the maxmium size of blob tags for single blob?

src/blob/utils/utils.ts
```diff
@@ -164,3 +165,79 @@ export function getUserDelegationKeyValue(
...
+  if (blobTagsString === '' || blobTagsString === undefined)
+  {
+    return undefined;
+  }
+  let blobTags:BlobTag[] = [];
+  const rawTags = blobTagsString.split("&");
+  rawTags.forEach((rawTag)=>{
+    const tagpair = rawTag.split("=");
+    blobTags.push({
+      key: decodeURIComponent(tagpair[0]),
```
XiaoningLiu, line 185: does server has same behavior on URI decode against tags?

XiaoningLiu (approved):
Approved with comments.

XiaoningLiu (approved):
Approved with comments.

src/blob/persistence/LokiBlobMetadataStore.ts
blueww: Good catch, it doesn't support access condition, except x-ms-if-tags.

src/blob/persistence/SqlBlobMetadataStore.ts
blueww: Per https://learn.microsoft.com/en-us/rest/api/storageservices/set-blob-tags?tabs=azure-ad#request-body , tag count <= 10, key length <= 128, value length  <=256
Add up the other chars like keyword, blanket, for 10 tags, will need < 250 chars.
So the total should be 4090, I change it to 4096

src/blob/utils/utils.ts
blueww: I have tested all allowed special chars that allowed in blob tag in test case, all works well.
The input chars is same as output chars.

When we get the input tag here, the special char is encoded, so we need decode it.

src/blob/persistence/LokiBlobMetadataStore.ts
```diff
@@ -3299,6 +3306,103 @@ export default class LokiBlobMetadataStore
...
+   * @returns {Promise<BlobTags | undefined>}
+   * @memberof LokiBlobMetadataStore
+   */
+  public async getBlobTag(
+    context: Context,
+    account: string,
+    container: string,
+    blob: string,
+    snapshot: string = "",
+    leaseAccessConditions: Models.LeaseAccessConditions | undefined,
```
EmmaZhu, line 3378: Does get tags operations require lease ID for a leased blob?

src/blob/utils/utils.ts
```diff
@@ -164,3 +165,81 @@ export function getUserDelegationKeyValue(
...
+    if (ContainsInvalidTagCharacter(tag.value)) {
+      throw StorageErrorFactory.getInvalidTag(contextID);
+    }
+  });
+}
+
+function ContainsInvalidTagCharacter(s: string): boolean{
+  for (let c of s)
+  {
+    if (!(c >= 'a' && c <= 'z' ||
```
EmmaZhu, line 229: Add embrace around (c >= 'a' && c <= 'z') to make the logic clearer.

tests/blob/apis/blob.test.ts
```diff
@@ -1327,6 +1328,472 @@ describe("BlobAPIs", () => {
...
+    const guid = "ca761232ed4211cebacd00aa0057b223";
+    const leaseClient = blockBlobClient.getBlobLeaseClient(guid);
+    await leaseClient.acquireLease(-1);
+
+    const tags = {
+      tag1: "val1",
+      tag2: "val2",
+    };
+    await blockBlobClient.setTags(tags, { conditions: { leaseId: leaseClient.leaseId } });
+    const response = await blockBlobClient.getTags({
```
EmmaZhu, line 1769: Should getTags still work without lease ID?

EmmaZhu (approved):

src/blob/persistence/LokiBlobMetadataStore.ts
blueww: Per the rest API doc, it does: https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-tags?tabs=azure-ad#request-headers

src/blob/utils/utils.ts
blueww: As we discussed, we agree with current code.

Issue #1929, opened 2023-04-20
Azurite Doesn't Handle Empty Table Fields Correctly When Quering With Not Equals


garrettbaski:
### Which service(blob, file, queue, table) does this issue concern?
Table

### Which version of the Azurite was used?
3.23.0

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
NPM

### What's the Node.js version?
v18.15.0

### What problem was encountered?
When querying tables, if you have an entity with a missing field and then query using something like ; 
```field != '' ```
That field will be returned in Azurite. However, from Azure Storage RP, it will not be returned.
Eg) If you add the two following items

Item 1 : {
PartitionKey : 1,
Row : 1,
Foo: "hi"
}

Item 2: {
PartitionKey: 1,
Row: 2
}

And then do a query for 
```Foo ne '' ```
You will get back both values in azurite but only the first value from storage. Aziure Storage seems to remove any entities where the field is not set when doing a query which removes it from being a viable candidate for these kinds of queries. 

### Steps to reproduce the issue?

Link to a test that will fail for Azurite but succeed for real Azure Storage can be found here : https://github.com/Azure/Azurite/pull/1930/files#diff-e0ef70376f6f826126897df2e43d8ab9fa613d813a176d0a443085500eb89734

### Have you found a mitigation/solution?
Yes, I have a PR, but the code is not great. But its at least an idea of how to fix it. PR Link here https://github.com/Azure/Azurite/pull/1930:

garrettbaski:
> @edwin-huber
> 
> I can repro this issue with storage explorer, do you have any idea how to fix it? Following is the Azurite debug log . (An entity without property "foo" can be query out with `foo ne ''`, in product Azure it won't be query out.)
> 
> ```
[...]

PR #1930, opened 2023-04-20
Issue 1929

garrettbaski:
This PR is probably too messy since I haven't contributed to this repo before and didn't have a must better way of adding this check with all of the Predicate/Token handling. But it should at least give the test cases and maybe some ideas on how to fix it to someone who knows the repo/language better.

See the issue here for some more info https://github.com/Azure/Azurite/issues/1929 but when querying table storage, if you have an entity with a not set field and you query on that field with a not equal, that entity will be returned since it is not equal. However this does not match the behavior in the real Azure Storage where if you query a specific field and that field is not set, it ignore that entity. 

For this PR, I basically added a check for any identifier to make sure the field is there doing 
``` item.properties.hasOwnProperty("<propertyValue">) ```
I'm not a fan of how it is jammed in. And there may be a better way to check the field, but I'm not super familiar with the code or typescript. 

The C# test tests the behavior. Current if ran against real Azure Storage, it will pass. However, if ran against Azurite it will fail. The added test in the table.entity.query.test.ts does the same thing

### CODE REVIEW

**Comment by garrettbaski** [comment]

**Comment by notheotherben** [comment]
Hi @garrettbaski,

Looking at this change, it looks like there is a ton of repetitive code - that's usually a sign that there's a better way to tackle the problem (either through an abstraction, or through a common injection point). Walking through the parser logic here, it seems like there are a number of steps happening, and I'll try to capture my understanding here so that if someone more familiar with the code disagrees, they can ideally point out where I've led you astray :).

1. The parser consists of a state machine which consumes the query tokens and proceeds to convert these into one of `Identifier`, `Operator` or `Value` tokens (themselves contained in a `TokenMap`).
2. In the `StatePredicateFinished` parser state, these tuples are converted into an `IPredicate` (such as the `BooleanPredicate`) which provides the `convertPredicateForLokiJS` function. This function essentially performs the transform from a "query string token map" into a "JavaScript token map" (and you'll see this being done with the `r/this.push(Value|Identifier|Operator)/` methods in each predicate implementation).
3. The `convertPredicateForLokiJS` function is actually executed in the `StateQueryFinished` parser state, which is responsible for converting the whole shebang into a `transcribedQuery` string (which should be valid JavaScript that can then be used to construct a `Function` for execution).

Right now, you've approached solving this problem by doing two things:
- You've introduced a condition within each `IPredicate` implementation which attempts to add `item.properties.hasOwnProperty(...)` at either the start or end of the predicate's "JavaScript token map" when `convertPredicateForLokiJS` is called.
- Assuming that there can only be one `Identifier` token within a predicate (this is not necessarily true, as we can assert `x eq y`).

Given my understanding of the way this library works, I think you may be better served by focusing on the 3rd portion of the translation (the `StateQueryFinished` state)  and in particular the `onProcess` method there - where each predicate is converted into its JavaScript form.

I believe introducing the following changeset there might be all that is necessary to implement support for this, as well as to ensure that we don't run into problems with evaluating `x eq y` either.

```diff --git a/src/table/persistence/QueryTranscriber/StateQueryFinished.ts b/src/table/persistence/QueryTranscriber/StateQueryFinished.ts
index 227d0b4..74ac024 100644
--- a/src/table/persistence/QueryTranscriber/StateQueryFinished.ts
+++ b/src/table/persistence/QueryTranscriber/StateQueryFinished.ts
@@ -26,6 +26,13 @@ export default class StateQueryFinished implements IQPState {
     for (const taggedPredicate of context.taggedPredicates) {
       let predicate = "";
       if (taggedPredicate !== undefined) {
+        const requiredIdentifiers = taggedPredicate.tokenMap.tokens.filter(token => token.type.isIdentifier() && token.token !== "**blena**");
+
+        for (const identifierToken of requiredIdentifiers) {
+          predicate += " ";
+          predicate += `item.properties.hasOwnProperty("${identifierToken.token.replace(/\\/g, "\\\\")}") &&`;
+        }
+
         const convertedPredicate = taggedPredicate.convertPredicateForLokiJS();
         for (const taggedPredicateToken of convertedPredicate.tokenMap.tokens) {
           predicate += " ";
```

Doing so will result in any predicate which references an `Identifier` being prefixed with `item.properties.hasOwnProperty($$identifier$$) &&`, for each identifier used, and it avoids the need for you to introduce any additional changes to the `IPredicates`, greatly simplifying your changeset. As a last caution - making the above change will cause your unit tests to fail as the ordering that you've currently implemented (either pre-fixing or post-fixing the condition) no longer applies, I think this is likely for the best as there's no value in evaluating the predicate if the field does not exist.

garrettbaski:
> Hi @garrettbaski,
> 
> Looking at this change, it looks like there is a ton of repetitive code - that's usually a sign that there's a better way to tackle the problem (either through an abstraction, or through a common injection point). Walking through the parser logic here, it seems like there are a number of steps happening, and I'll try to capture my understanding here so that if someone more familiar with the code disagrees, they can ideally point out where I've led you astray :).
> 
> 1. The parser consists of a state machine which consumes the query tokens and proceeds to convert these into one of `Identifier`, `Operator` or `Value` tokens (themselves contained in a `TokenMap`).
> 2. In the `StatePredicateFinished` parser state, these tuples are converted into an `IPredicate` (such as the `BooleanPredicate`) which provides the `convertPredicateForLokiJS` function. This function essentially performs the transform from a "query string token map" into a "JavaScript token map" (and you'll see this being done with the `r/this.push(Value|Identifier|Operator)/` methods in each predicate implementation).
> 3. The `convertPredicateForLokiJS` function is actually executed in the `StateQueryFinished` parser state, which is responsible for converting the whole shebang into a `transcribedQuery` string (which should be valid JavaScript that can then be used to construct a `Function` for execution).
> 
> Right now, you've approached solving this problem by doing two things:
> 
> * You've introduced a condition within each `IPredicate` implementation which attempts to add `item.properties.hasOwnProperty(...)` at either the start or end of the predicate's "JavaScript token map" when `convertPredicateForLokiJS` is called.
> * Assuming that there can only be one `Identifier` token within a predicate (this is not necessarily true, as we can assert `x eq y`).
> 
> Given my understanding of the way this library works, I think you may be better served by focusing on the 3rd portion of the translation (the `StateQueryFinished` state) and in particular the `onProcess` method there - where each predicate is converted into its JavaScript form.
> 
> I believe introducing the following changeset there might be all that is necessary to implement support for this, as well as to ensure that we don't run into problems with evaluating `x eq y` either.
> 
> ```diff
> index 227d0b4..74ac024 100644
> --- a/src/table/persistence/QueryTranscriber/StateQueryFinished.ts
> +++ b/src/table/persistence/QueryTranscriber/StateQueryFinished.ts
> @@ -26,6 +26,13 @@ export default class StateQueryFinished implements IQPState {
>      for (const taggedPredicate of context.taggedPredicates) {
>        let predicate = "";
>        if (taggedPredicate !== undefined) {
> +        const requiredIdentifiers = taggedPredicate.tokenMap.tokens.filter(token => token.type.isIdentifier() && token.token !== "**blena**");
> +
> +        for (const identifierToken of requiredIdentifiers) {
> +          predicate += " ";
> +          predicate += `item.properties.hasOwnProperty("${identifierToken.token.replace(/\\/g, "\\\\")}") &&`;
> +        }
> +
>          const convertedPredicate = taggedPredicate.convertPredicateForLokiJS();
>          for (const taggedPredicateToken of convertedPredicate.tokenMap.tokens) {
>            predicate += " ";
> ```
> 
> Doing so will result in any predicate which references an `Identifier` being prefixed with `item.properties.hasOwnProperty($$identifier$$) &&`, for each identifier used, and it avoids the need for you to introduce any additional changes to the `IPredicates`, greatly simplifying your changeset. As a last caution - making the above change will cause your unit tests to fail as the ordering that you've currently implemented (either pre-fixing or post-fixing the condition) no longer applies, I think this is likely for the best as there's no value in evaluating the predicate if the field does not exist.

Yeah this is a much better way of doing it. I couldn't think of how to do it in a single spot. I changed it to this and fixed up all the tests.

notheotherben:
Looks solid to me, thanks Garrett - hopefully this is a bit quicker for @edwin-huber to review now, shout if there's anything else I can help with (unfortunately being an approver is not one of those things right now).

notheotherben (approved):

blueww:
@edwin-huber , @XiaoningLiu 
Would you please help to review if the fix is good to you?

garrettbaski:
I think this will be invalidated by https://github.com/Azure/Azurite/pull/1975 if that is merged first. In that case, I just need to ensure that the test case that I have here successfully executes and if so then this will not be needed.

blueww:
@garrettbaski 

After https://github.com/Azure/Azurite/pull/1975 merged, if the issue of this PR still exist?
If so, would you please update this PR based on the current code base?

Thanks for your contribution!

blueww (approved):
Only test case change. Looks good to me.

blueww:
@XiaoningLiu  , @EmmaZhu 
Any comments for this PR?

XiaoningLiu (approved):

blueww (approved):

Issue #1972, opened 2023-05-22
[Table] Filter condition fails if the query contains a backtick `


garrettbaski:
Which service(blob, file, queue, table) does this issue concern?
Table

Which version of the Azurite was used?
3.23.0

Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
NPM

What's the Node.js version?
v18.15.0

What problem was encountered?
If your query has a backtick, it will return a 500 saying "The query condition specified in the request is invalid". It looks like in the string predicate the code says : 

```     
     taggedToken.token =
        "`" +
        // need to convert double apostrope to single
        this.replaceDoubleApostrophes(
          taggedToken.token.substring(1, taggedToken.token.length - 1)
        ) +
        "`";
```

Because of this, it looks like having a backtick in the query basically escapes the query. 

Adding the test below to the table.entity.query.tests.ts reproduces this

``` 
  it("20. should work when getting special characters, @loki", async () => {
    const tableClient = createAzureDataTablesClient(
      testLocalAzuriteInstance,
      getUniqueName("specialcharactercheck")
    );

    await tableClient.createTable();

    const partitionKey = createUniquePartitionKey("");

    // Foo has some special characters
    const result1 = await tableClient.createEntity({
      partitionKey: partitionKey,
      rowKey: `1`,
      foo: "TestVal`",
    });
    assert.notStrictEqual(result1.etag, undefined);

    const maxPageSize = 5;
    const entities = tableClient.listEntities<TableEntity<{ foo: string }>>({
      queryOptions: {
        filter: `PartitionKey eq '${partitionKey}' and foo eq 'TestVal\`'`
      }
    });
    let all: TableEntity<{ foo: string }>[] = [];
    for await (const entity of entities.byPage({
      maxPageSize
    })) {
      all = [...all, ...entity];
    }
    assert.strictEqual(all.length, 1);

    await tableClient.deleteTable();
  });
```

Have you found a mitigation/solution?
Not yet not

blueww:
@edwin-huber 

It looks the failure is realted with the code change from your commit https://github.com/Azure/Azurite/commit/7dd63170230c034ce01116185decc122748fd74c

Would you please help to look?

blueww:
As #1973 is merged, the fix will be out in next Azurite release.

PR #1973, opened 2023-05-23
fix: Resolve issues with reserved character use in query strings (allowing both interpolation injection and causing evaluation failures)

notheotherben:
This PR makes some changes to the way that strings are converted into their LokiJS representations with the goal of addressing problems with the use of both backticks (<code>\`</code>) as well as the potential for query string injection through the use of `${interpolations}`.

It accounts for the fix introduced in #1788 for the use of backslashes (`\`), which was missing a repro test (this has been introduced) and I've added test cases to ensure we don't regress on the injection risks here.


blueww:
@notheotherben 

Thanks for raising this PR!
It  can fix the issue: https://github.com/Azure/Azurite/issues/1972

@edwin-huber , @XiaoningLiu , @EmmaZhu 
Would you please help to review?
Do you have any comments for this change?

edwin-huber:
Reviewing

edwin-huber (approved):

blueww (approved):

PR #1975, opened 2023-05-24
feat: Introduce a new Table query interpreter which removes the ability to conduct query injection

notheotherben:
This PR is a ground-up rewrite of the query parsing functionality within Azurite's Table server implementation. The goal here is to do the following:

1. Make this substantially easier to review and work with (by moving to a simple recursive descent parser).
2. Replace the current `new Function(...)` based query translation model (which opens avenues for query injection) with a limited interpreter which endeavours to block this behaviour.
3. Expand the test suite for the parser to more explicitly outline what we're testing and make spotting issues faster.

This PR is a follow-up on #1973 which goes a step beyond the simplistic escape-defences introduced there and ensures that the actual evaluation of queries occurs in a deterministic sandbox which does not have the ability to execute arbitrary code.

I acknowledge that this is a significant PR, so please shout if there's anything I can do to help make it easier to review, or if you'd like to sit down and talk through it together (you've all got my contact details on Teams).


src/table/persistence/QueryInterpreter/QueryNodes/NotEqualsNode.ts
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    return this.left.evaluate(context) !== this.right.evaluate(context)
+  }
+
+  private backwardsCompatibleGuidEvaluate(context: IQueryContext): boolean {
+    const left = this.left instanceof GuidNode ? this.left.stringGuid() : this.left.evaluate(context)
+    const right = this.right instanceof GuidNode ? this.right.stringGuid() : this.right.evaluate(context)
+
+    return left !== right
```
blueww, line 22: This issue still exist: https://github.com/Azure/Azurite/issues/1929
This code change might can fix it. And the `evaluate()` function above should also be fix similarly.
```suggestion
    if (left === undefined || right === undefined)
    {
      return false;
    }
    return left !== right
```

src/table/persistence/QueryInterpreter/QueryNodes/ExpressionNode.ts
```diff
@@ -0,0 +1,18 @@
+import { IQueryContext } from "../IQueryContext";
+import IQueryNode from "./IQueryNode";
+
+export default class ExpressionNode implements IQueryNode {
```
blueww, line 4: Could we add some comments/samples to describe each kind of node usage, to help the code understanding.

src/table/persistence/QueryInterpreter/QueryNodes/NotEqualsNode.ts
notheotherben: Thank you, I've added that improvement as well as a test case to ensure we don't regress on this.

src/table/persistence/QueryInterpreter/QueryNodes/NotEqualsNode.ts
```diff
@@ -2,22 +2,31 @@ import { IQueryContext } from "../IQueryContext";
...
-    if (!this.backwardsCompatibleGuidEvaluate(context)) {
-      return false;
-    }
-
     const left = this.left.evaluate(context);
     const right = this.right.evaluate(context);
 
     // If either side is undefined, we should not match - this only occurs in scenarios where
     // the field itself doesn't exist on the entity.
-    return left !== right && left !== undefined && right !== undefined;
```
blueww: (Optional) Better add a test case for query with not exist property != value. 
Or I can add a test case after the PR is merged.

blueww (approved):
The PR looks good to me. 
I will sync with Xiaoning and Emma offline and see if we can merge it now.

src/table/persistence/QueryInterpreter/IQueryContext.ts
```diff
@@ -0,0 +1,23 @@
...
+export type IQueryContext = Entity | Table;
+
+/**
+ * Determines whether a given query context contains a table entity or not.
+ * 
+ * @param {IQueryContext} context The context which should be tested.
+ * @returns {boolean}
+ */
+export function isEntity(context: IQueryContext): context is Entity {
+  return !!(<Entity>context).PartitionKey
```
XiaoningLiu, line 12: Missing ending `;`, needs to check other code changes.

src/table/persistence/QueryInterpreter/QueryLexer.ts
```diff
@@ -0,0 +1,195 @@
...
+}
+
+/**
+ * This is the lexer responsible for converting a query string into a stream of tokens.
+ * These tokens are logical chunks of the query which can then be consumed by the parser,
+ * while not technically necessary, this does help simplify the parsing logic (allowing it
+ * to operate on the higher-level tokens directly).
+ */
+export class QueryLexer {
+  constructor(private query: string) { }
```
XiaoningLiu, line 24: Azurite is using upper case for 1st character of member functions. Needs to align with existing Azurite naming practice.

notheotherben: Can you show me an example of where this is done? I've looked through a number of classes within the `table` section, as well as others and not found any cases where `TitleCase` is used for method names (either within classes or in ambient functions).

(Probably not a bad idea to add a linting preset to enforce this as well so that others don't run into the same problem 😄)

src/table/persistence/QueryInterpreter/IQueryContext.ts
notheotherben: Thanks, I've done another pass through looking for any outliers here and I think I've got them all.

src/table/persistence/QueryInterpreter/QueryLexer.ts
XiaoningLiu: Ah, you are right. We should bring back our linter rules.

EmmaZhu (approved):

PR #2073, opened 2023-07-26
Refactor Table Batch

edwin-huber:
Refactoring started in Table Batch to address some of the outstanding issues.

Moving interaction with persistence layer to a repository class.
Corrected a test for Query logic in REST Batch Tests


edwin-huber:
Moving to a larger refactoring.

Issue #2078, opened 2023-07-27
Query on table with empty partition keys doesn't return any result after upgrade to 3.25.0 


mariusz-rubin:
### Which service(blob, file, queue, table) does this issue concern?
Table

### Which version of the Azurite was used?
3.25.0

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
npm

### What's the Node.js version?
20.2.0

### What problem was encountered?
Executing `QueryAsync` on `TableClient` doesn't return the expected entity when `PartitionKey` is set to an empty string.
This works with the previous Azurite version 3.24.0.
Using Nuget package `Azure.Data.Tables` in version 12.8.0

### Steps to reproduce the issue?
The C# code below is inserting an entity with empty `PartitionKey`.
Getting the entity by `PartitionKey` and `RowKey` works but filtering with `QueryAsync` by a text property doesn't return any result.
```csharp
using Azure;
using Azure.Data.Tables;

var tableClient = new TableClient(
    connectionString: "UseDevelopmentStorage=true",
    tableName: "Test");

await tableClient.CreateIfNotExistsAsync();

try
{
    const string textValue = "Test value";

    var entity = new Entity
    {
        PartitionKey = string.Empty,
        RowKey = Guid.NewGuid().ToString(),
        Text = textValue
    };

    await tableClient.AddEntityAsync(entity);

    var gotEntity = await tableClient.GetEntityIfExistsAsync<Entity>(entity.PartitionKey, entity.RowKey);
    Console.WriteLine($"Got entity: '{gotEntity.HasValue}'");

    var foundEntities = await tableClient.QueryAsync<Entity>(x => x.Text == textValue).ToListAsync();
    Console.WriteLine($"Found entity: '{foundEntities.Any()}'");
}
finally
{
    await tableClient.DeleteAsync();
}

class Entity : ITableEntity
{
    public string PartitionKey { get; set; } = string.Empty;
    public string RowKey { get; set; } = string.Empty;
    public string Text { get; set; } = string.Empty;
    public DateTimeOffset? Timestamp { get; set; }
    public ETag ETag { get; set; }
}
```
Referenced packages:
```
 <ItemGroup>
    <PackageReference Include="Azure.Data.Tables" Version="12.8.0" />
    <PackageReference Include="System.Linq.Async" Version="6.0.1" />
  </ItemGroup>
```

### Have you found a mitigation/solution?
Using the previous Azurite version 3.24.0.

[...]

PR #2079, opened 2023-07-27
fix : Resolve issues with empty partition key query not returning anything

garrettbaski:
Fixing issue #2078. The issue was caused by checking if the partition key existed by doing !!<object>.PartitionKey. In JS this resolves to false if the value is an empty string. Instead, do a "hasOwnProperty" on the partition key to check if the field exists or not instead of checking if the value exist to solve this problem.


src/table/persistence/QueryInterpreter/IQueryContext.ts
```diff
@@ -9,7 +9,7 @@ export type IQueryContext = Entity | Table;
  * @returns {boolean}
  */
 export function isEntity(context: IQueryContext): context is Entity {
-  return !!(<Entity>context).PartitionKey;
+  return !!(<Entity>context).hasOwnProperty('PartitionKey');
```
notheotherben, line 12: optional nit: I think the leading `!!` is redundant here as `.hasOwnProperty` is already a boolean value.

notheotherben (approved):
Looks good to me, thank you @garrettbaski for getting this sorted.

XiaoningLiu:
Thanks @garrettbaski! Looks good, and great if able to further update per @notheotherben's suggestion.

notheotherben:
(for what it's worth, my suggestion isn't a functional blocker, just a stylistic one)

src/table/persistence/QueryInterpreter/IQueryContext.ts
garrettbaski: I went ahead and removed the ```!!```

EmmaZhu (approved):

blueww (approved):

PR #2113, opened 2023-08-16
Fix validation of Blob SAS token when using the second key for an account in `AZURITE_ACCOUNTS`

stekycz:
This PR fixes use cases below when using `AZURITE_ACCOUNTS` defining a second key for the account.

- A blob request using any key but any of the 2 keys defined in `AZURITE_ACCOUNTS` passes instead of failing.
- A blob request using the second key defined in `AZURITE_ACCOUNTS` fails instead of passing.

### DIFF
--- a/tests/blob/sas.test.ts
+++ b/tests/blob/sas.test.ts
@@ -30,6 +30,8 @@ const EMULATOR_ACCOUNT2_NAME = "devstoreaccount2";
 const EMULATOR_ACCOUNT2_KEY_STR =
   "MTAwCjE2NQoyMjUKMTAzCjIxOAoyNDEKNDAKNzgKMTkxCjE3OAoyMTQKMTY5CjIxMwo2MQoyNTIKMTQxCg==";
 
+const EMULATOR_ACCOUNT_KEY2_STR = "testing_key";
+
 // Set true to enable debug log
 configLogger(false);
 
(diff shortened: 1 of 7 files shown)


tests/blob/sas.test.ts
```diff
@@ -58,6 +60,36 @@ describe("Shared Access Signature (SAS) authentication", () => {
     )
   );
 
+  const serviceClient1 = new BlobServiceClient(
```
blueww, line 63: It would be better to give a more specific name  than "serviceClient1".

blueww (approved):
Thanks for the good catch!

It would be better if you could also help to fix other place has the similar issue:

https://github.com/Azure/Azurite/blob/1924903baa9e78ee7726cdcb1c8fdead3d14e1f6/src/blob/authentication/BlobSASAuthenticator.ts#L279
https://github.com/Azure/Azurite/blob/1924903baa9e78ee7726cdcb1c8fdead3d14e1f6/src/table/authentication/TableSASAuthenticator.ts#L158
https://github.com/Azure/Azurite/blob/1924903baa9e78ee7726cdcb1c8fdead3d14e1f6/src/table/authentication/AccountSASAuthenticator.ts#L128
https://github.com/Azure/Azurite/blob/1924903baa9e78ee7726cdcb1c8fdead3d14e1f6/src/queue/authentication/QueueSASAuthenticator.ts#L157

tests/blob/sas.test.ts
stekycz: Sure thing, I am just pushing the change

stekycz:
I have addressed the comments above. I did not add any new regression tests for other 4 authenticators though. I believe it is good enough to have it covered once but let me know if you want me to add all those tests.

blueww:
@stekycz 

Thanks so much for the fix!

It would be great if you can also added test cases for the other fix place!
However, if adding the test cases is too high cost, we can also add it later.

stekycz:
@blueww I would prefer adding tests later. This fix already unblocks usage for specific test cases for me.

I am not sure where those tests should be added. Test structure differs a lot among blob, table and queue storage kinds.

XiaoningLiu (approved):

PR #2125, opened 2023-08-21
Auto generate from swagger

EmmaZhu:
Updated blob swagger spec, now can auto generate interfaces from swagger spec, no manual change needed anymore.


src/blob/generated/handlers/IServiceHandler.ts
```diff
@@ -19,6 +19,7 @@ export default interface IServiceHandler {
   listContainersSegment(options: Models.ServiceListContainersSegmentOptionalParams, context: Context): Promise<Models.ServiceListContainersSegmentResponse>;
   getUserDelegationKey(keyInfo: Models.KeyInfo, options: Models.ServiceGetUserDelegationKeyOptionalParams, context: Context): Promise<Models.ServiceGetUserDelegationKeyResponse>;
   getAccountInfo(context: Context): Promise<Models.ServiceGetAccountInfoResponse>;
+  getAccountInfoWithHead(context: Context): Promise<Models.ServiceGetAccountInfoWithHeadResponse>;
```
blueww, line 22: What's the function of "getAccountInfoWithHead" API?

EmmaZhu: It just calls getAccountInfo method. You can find following code in BlobHandler.ts.
```
  public async getAccountInfoWithHead(
    context: Context
  ): Promise<Models.BlobGetAccountInfoResponse> {
    return this.getAccountInfo(context);
  }
```

blueww (approved):
Just to confirm, after this PR merge, the generated code should be aligned with the current code in generated folder without manual change.

PR #2165, opened 2023-09-13
Adds NUnit Tests for Table Storage

edwin-huber:
This PR adds NUnit Tests for Table Storage to help reliably repro issues from DotNet developers and SDKs which behave differently to the Node and Javascript SDKs.

These tests will then be added to the Azure Pipeline for PRs to ensure that we have regression testing and can be extended to cover additional properties and behaviors which are relevant in those SDKs.

To test, simply start Azurite Table Storage using "npm run table", and the open the Test Solution under tests/table/dotnet

You can now test and debug DotNet repros and issues in parallel using Visual Studio and VSCode.


blueww (approved):
The change looks good to me:
1. All the test cases in the removed files are added back in the modified/new files
2. I have run the test case locally with Azurite , and all passed.

@XiaoningLiu , would you please see if any comments?

XiaoningLiu (approved):

Issue #2169, opened 2023-09-15
Guid queries on table returns all rows without value


FranceyD:
### Which service(blob, file, queue, table) does this issue concern?
Table

### Which version of the Azurite was used?
3.25, 3.25.1 or 3.26. No problem with Azurite 3.24 or lower

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
DockerHub and npm

### What's the Node.js version?
14.21.3

### What problem was encountered?
Doing a query with a specified guid returns all rows with the specified guid, but also all rows where there is no value in the column.
On Table storage and on Azurite < 3.25, only the rows with the specified guid were returned.

### Steps to reproduce the issue?
Create a table. Add some rows only with a partition key and row key. Do a query on any empty column with any guid like this: `AnyColumn eq guid'43a29cc9-cc32-4f6c-a40b-1ada7f4bc0b0'`. You receive all rows of the table.

If possible, please provide the debug log using the -d parameter, replacing \<pathtodebuglog\> with an appropriate path for your OS, or review the instructions for docker containers:

```
-d "<pathtodebuglog>"
```

Please be sure to remove any PII or sensitive information before sharing!  
The debug log will log raw request headers and bodies, so that we can replay these against Azurite using REST and create tests to validate resolution.

### Have you found a mitigation/solution?
Yes, using Azurite 3.24

Edit: I Fixed an error with Azurite version. The issue appeared with Azurite 3.25, not Azurite 3.25.1

blueww:
@notheotherben

I can repro this issue. 
It looks the issue related with the recent [table query PR](https://github.com/Azure/Azurite/pull/2077) from you.

The problem looks in function `compare()`: https://github.com/Azure/Azurite/blob/fa485cd10f6c4f407be915cc16810d6e52c015bb/src/table/persistence/QueryInterpreter/QueryNodes/GuidNode.ts#L25
When "otherValue" is undefined, "thisValue < otherValue" and "thisValue > otherValue" both are not true, so finally 0 will be returned which means "thisValue" equals to "otherValue", which is also not correct.
However, when "otherValue" is undefined, it looks return -1, 0, 1 in function `compare()` are all wrong (since all comparison should fail in this case, include eq, gt, lt, ge, le, ne). So not sure how to fix this issue, looks any return value from the function `compare()` is not correct.
[...]

PR #2173, opened 2023-09-18
fix(table): Ensure that we correctly handle value comparisons against null/empty/undefined values for GUIDs and DateTimes

notheotherben:
This PR fixes the issues raised in #2169 which pertain to the comparison between empty (and by extension, null and missing) properties and GUIDs. This issue was reproduced with the included test suite additions, and fixed by handling the `null`/`undefined`/empty case explicitly for GUID comparisons.

### DIFF
--- a/src/table/persistence/QueryInterpreter/QueryNodes/DateTimeNode.ts
+++ b/src/table/persistence/QueryInterpreter/QueryNodes/DateTimeNode.ts
@@ -16,22 +16,31 @@ export default class DateTimeNode<T> extends ValueNode {
   }
 
   compare(context: IQueryContext, other: IQueryNode): number {
+    const otherValue = other.evaluate(context);
+
+    // NOTE(notheotherben): This is a special case for the `null` value, which is not a valid datetime value in Azure Storage
+    //                      but is considered a valid input for "epoch" in the JS Date constructor. We're explicitly handling
+    //                      returning NaN here to ensure that null doesn't match dates in the table.
+    if (this.value === null || otherValue === null) {
+      return NaN;
+    }
+
     // NOTE(notheotherben): This approach leverages the fact that the `Date` constructor will parse ISO8601 strings
     //                 however it runs into a limitation of the accuracy of JS dates (which are limited to millisecond
     //                 resolution). As a result, we're effectively truncating the value to millisecond precision by doing
     //                 this. This is fundamentally a trade-off between enforcing valid datetime values and providing perfect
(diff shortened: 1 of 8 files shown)


tests/table/unit/query.interpreter.unit.test.ts
```diff
@@ -269,6 +270,21 @@ describe("Query Interpreter", () => {
...
+      },
+      {
+        name: " GUID compare against null property",
+        originalQuery: "nullValue eq guid'00000000-0000-0000-0000-000000000000'",
+        expectedResult: false
+      },
+      {
+        name: "GUID compare against empty string",
+        originalQuery: "emptyString eq guid'00000000-0000-0000-0000-000000000000'",
+        expectedResult: false
```
blueww, line 287: The PR generally looks good to me. 
However, we might should also add some test case for gt, lt, ge, le, ne , to make sure all kinds of comparison will return false when the column to compare not exist.

src/table/persistence/QueryInterpreter/QueryNodes/DateTimeNode.ts
```diff
@@ -25,7 +25,9 @@ export default class DateTimeNode<T> extends ValueNode {
     const thisValue = new Date(this.value);
     const otherValue = new Date(other.evaluate(context));
 
-    if (thisValue.valueOf() < otherValue.valueOf()) {
+    if (isNaN(thisValue.valueOf()) || isNaN(otherValue.valueOf())) {
```
XiaoningLiu, line 28: Do we need this check? If any value is `NaN` following 3 checks should be all `false`. Correct me if wrong.

notheotherben: We don't strictly need it, you're spot on - my thoughts were that by being explicit here we're far less likely to have someone make changes to the logic in future and break this behaviour (consider this self documenting code).

tests/table/unit/query.interpreter.unit.test.ts
notheotherben: Let me see if I can put together a generic harness for validating this stuff :+1:

notheotherben: PR has been updated with a new set of tests to cover this (+some additional changes to the handling of `ne` and `binary` to make those a bit more robust per the new test constraints).

blueww (approved):

XiaoningLiu (approved):

PR #2221, opened 2023-10-13
Add some development options to SQL metastore

mahmoudbahaa:
Add some development/debugging options to SQL metastore:

1. ```cleardb``` that clear the database at start which is only set to true in test server
2. ```soft delete``` option that can be used to debug GC problems

### DIFF
--- a/src/blob/persistence/SqlBlobMetadataStore.ts
+++ b/src/blob/persistence/SqlBlobMetadataStore.ts
@@ -108,7 +108,9 @@ export default class SqlBlobMetadataStore implements IBlobMetadataStore {
    */
   public constructor(
     connectionURI: string,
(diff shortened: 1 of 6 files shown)


src/blob/persistence/SqlBlobMetadataStore.ts
```diff
@@ -365,8 +367,12 @@ export default class SqlBlobMetadataStore implements IBlobMetadataStore {
       }
     );
 
-    // TODO: sync() is only for development purpose, use migration for production
-    await this.sequelize.sync();
+    if (this.clearDB) {
```
blueww, line 370: Blob content is on local disk, so clear DB will make data not aligned.

Issue #2224, opened 2023-10-14
Add missing functionality in SQL based metadata implementation. to be in sync with file based one.


mahmoudbahaa:
Currently Blob Copy & Page Blob are not supported by SQL based metadata implementation.

Need to make both SQL based metadata implementation support all functionality in File based metadata implementation

PR #2225, opened 2023-10-14
Add Blob Copy & Page Blob support to by SQL based metadata implementation in Blob API and Sync Loki and SQL metadata stores

mahmoudbahaa:
Add Blob Copy & Page Blob support to by SQL based metadata implementation in Blob API. This makes the SQL based metadata implementation in sync with File based one. (issue #2224)

I divided the PR into very small mini commits for better visibility but most of these commits are small either in scope or code.
 
Changes are as follows:

In code: ([SqlBlobMetadataStore.ts])

1. Changed properties from separate fields into a single one to allow serializing de-serializing of remaining properties as well.
2. Added missing await from some async functions in SqlBlobMetadataStore
3. Ported ```copyFromURL``` from Loki to SQL blobMeatadataStore
4. Ported ```appendBlock``` from Loki to SQL blobMeatadataStore
5. Ported ```clearRange``` from Loki to SQL blobMeatadataStore
6. Ported ```getPageRanges``` from Loki to SQL blobMeatadataStore
7. Ported ```resizePageBlob``` from Loki to SQL blobMeatadataStore
8. Ported ```updateSequenceNumber``` from Loki to SQL blobMeatadataStore
9. Return blobCommittedBlockCount for append blob in getBlobProperties in SQL blobMetadataStore
10. Check for CopyIfExists condition in startCopyFromURL in SQL blobMetadaStore

In Tests:

1. Add @sql to all @loki tests
2. Ignore ```generateAccountSASQueryParameters``` in sas blob tests as it depend on timezone and fails even then and test another component not Azurite
3. Skip production style URL test as it fails in new sdk //To be fixed
4. Remove skipping of "list uncimmited blob from container"  in blob api container test since now it is working
5. Added await to 2 async functions in 2 tests that was causing these 2 tests to be flaky

Others:

1. Changed readme and changelog to relfect sync btween Loki and SQL blob metadata stores.

Note: I recommend removing the @loki and @sql all together and run all test cases without grep for both loki and sql to avoid descync in the future but its your call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -488,8 +488,6 @@ This feature is in preview, when Azurite changes database table schema, you need
 
 > Note. Need to manually create database before starting Azurite instance.
 
-> Note. Blob Copy & Page Blob are not supported by SQL based metadata implementation.
-
 > Tips. Create database instance quickly with docker, for example `docker run --name mysql -p 3306:3306 -e MYSQL_ROOT_PASSWORD=my-secret-pw -d mysql:latest`. Grant external access and create database `azurite_blob` using `docker exec mysql mysql -u root -pmy-secret-pw -e "GRANT ALL PRIVILEGES ON *.* TO 'root'@'%' WITH GRANT OPTION; FLUSH PRIVILEGES; create database azurite_blob;"`. Notice that, above commands are examples, you need to carefully define the access permissions in your production environment.
 
 ## HTTPS Setup
--- a/src/blob/persistence/SqlBlobMetadataStore.ts
+++ b/src/blob/persistence/SqlBlobMetadataStore.ts
@@ -2632,14 +2632,190 @@ export default class SqlBlobMetadataStore implements IBlobMetadataStore {
     });
   }
 
-  public copyFromURL(
+  public async copyFromURL(
     context: Context,
     source: BlobId,
     destination: BlobId,
     copySource: string,
-    metadata: Models.BlobMetadata | undefined
+    metadata: Models.BlobMetadata | undefined,
+    tier: Models.AccessTier | undefined,
+    options: Models.BlobCopyFromURLOptionalParams = {}
   ): Promise<Models.BlobPropertiesInternal> {
-    throw new Error("Method not implemented.");
+    return await this.sequelize.transaction(async (t) => {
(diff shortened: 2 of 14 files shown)


tests/blob/specialnaming.test.ts
```diff
@@ -445,7 +445,9 @@ describe("SpecialNaming", () => {
     assert.notDeepEqual(response.segment.blobItems.length, 0);
   });
 
-  it(`Should work with production style URL when ${productionStyleHostName} is resolvable`, async () => {
+  // This now doesn't work (Error: Unable to extract accountName with provided information.).
+  // As the sdk expect an url in the form devstoreaccount1.blob.localhost instead of devstoreaccount1.localhost
+  it.skip(`Should work with production style URL when ${productionStyleHostName} is resolvable`, async () => {
```
blueww, line 450: This looks already won't run as don't have loki or sql tag, so looks don't need skip it.

README.md
```diff
@@ -488,8 +488,6 @@ This feature is in preview, when Azurite changes database table schema, you need
 
 > Note. Need to manually create database before starting Azurite instance.
 
-> Note. Blob Copy & Page Blob are not supported by SQL based metadata implementation.
-
```
blueww, line 492: The support matrix section in ReadMe.md should also be updated.
https://github.com/Azure/Azurite/blob/9208664c196acb49b133f6152296c7770c0a1d0e/README.md?plain=1#L1011

src/blob/persistence/SqlBlobMetadataStore.ts
```diff
@@ -238,37 +235,8 @@ export default class SqlBlobMetadataStore implements IBlobMetadataStore {
...
-        blobSequenceNumber: {
-          type: "VARCHAR(63)"
-        },
-        accessTier: {
-          type: "VARCHAR(31)"
-        },
-        contentProperties: {
-          type: "VARCHAR(1023)"
+        properties: {
+          type: "VARCHAR(4095)"
```
blueww, line 239: Will this cause the saved blobs in old Azurite version, can't be read out with new Azurite version?

Issue #1332, opened 2022-02-09
Enable Azurite to cleanup data files used for persistence 


edwin-huber:
This issue is created to track the work to enable the cleanup of data files uses for persistence.
Requirements:

- [ ] 1. Needs to be compatible and configurable as VS Code Extension
- [ ] 2. Needs to be compatible and configurable with Visual Studio 2022
- [ ] 3. Needs to support no-persistence mode (best performance)
- [ ] 4. Needs to support partitioning files on partitionKey (see #1331)

Optional:
- [ ] 1. Should allow option to preserve or snapshot / backup persistent storage
- [ ] 2. Should unify the persistence format, naming and locations across all Azurite Services

edwin-huber:
Dependency on #1061

Issue #1331, opened 2022-02-09
Improve Table Service Performance


edwin-huber:
I am actively looking at options to improve performance.
The best option I have found until we check in a change for this is to change the the constructor for LokiTableMetadataStore in  ```src\table\persistence\LokiTableMetadataStore.ts```

You have 2 options.

Turn off persistence:

```typescript
  public constructor(public readonly lokiDBPath: string) {
       this.db = new Loki(lokiDBPath, {
    });
  }
```

Enable the lfsa adapter, which creates files per partition:

```typescript
  public constructor(public readonly lokiDBPath: string) {
    const lfsa = require("../../../node_modules/lokijs/src/loki-fs-structured-adapter.js");
    const lfsaAdapter = new lfsa();
    this.db = new Loki(lokiDBPath, {
      autosave: true,
      adapter: lfsaAdapter
      autosaveInterval: 5000
    });
  }
```

Addresses #1198

edwin-huber:
Dependency on #1061

joelverhagen:
@edwin-huber, from my testing Table didn't get faster in-memory. The slow path for blob and queue was writing extents which is not done at all in tables. LokiJS does not persist immediately upon write so Loki + Table doesn't have disk in the request context, just in a background persist every 5 seconds. Perhaps I am misunderstanding the original issue however.

joelverhagen:
I wrote a blog about my change which describes the perf change: https://www.joelverhagen.com/blog/2023/11/azurite-in-memory

edwin-huber:
> a background persist every 5 seconds. 

It was exactly this IO causing systems to hang, writing large amounts of table data to disk...

joelverhagen:
Makes sense, thanks for clarifying this for me !

Issue #2227, opened 2023-10-15
[Feature request] Allow disk-based persistence to be completely disabled


joelverhagen:
### Which service(blob, file, queue, table) does this issue concern?

all

### Which version of the Azurite was used?

3.26.0

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)

VS Code, npm

### What's the Node.js version?

18.18.2

### What problem was encountered?

Currently, there are two paths that are hard-coded to use disk-based persistence: LokiJS and extent storage. It would be great to enable an `--inMemoryPersistence` option that makes all data (including extents) stored only in memory.

### Steps to reproduce the issue?

All 3 service types (blob, queue, table) use LokiJS which periodically saves to disk. Queue and Blob use extent storage on disk.

### Have you found a mitigation/solution?

No

joelverhagen:
I've started a draft PR here: https://github.com/Azure/Azurite/pull/2228. Let me know if this is something you'd be interested in accepting! Most of the code changes were related to plumbing and documenting configs :)

blueww:
Close as PR @2228 is merged, will in the next Azurite release, which target early/middle of Dec.

PR #2228, opened 2023-10-15
Allow disk-based persistence to be completely disabled

joelverhagen:
### PR Branch Destination

This adds a new `--inMemoryStorage` option which allows the disk-based extent storage to become memory based and the LokiJS persistence settings to use only memory.

I'm happy to break this up into smaller PRs if you'd like.

This addresses the following feature request (also by me): https://github.com/Azure/Azurite/issues/2227

Summary of changes:
- Add design spec for this feature. I added a new `docs` structure mimicking how we do public specs on NuGet team.
- Introduce `--inMemoryStorage` option and plumb it through the code
- Add `MemoryExtentStore` as an alternative to `FSExtentStore`
- Fix some flaky tests (flaky prior to my change, judging by the AzDO CI build history)
- Refactor `QueueServer` and `TableServer` creation to a test factory, allowing centralized configuration.
- Add steps to the ADO build pipeline to run all tests with `--inMemoryStorage` right after disk-based.

### Always Add Test Cases

All test cases (except 1) work with the `--inMemoryPersistence` option. The one test failing is skipped for `--inMemoryPersistence`. It makes sense that it fails because it requires reading a checked-in legacy LokiJS disk persistence.

### Design spec

The render design spec is avalable here: https://github.com/joelverhagen/Azurite/blob/in-memory/docs/designs/2023-10-in-memory-persistence.md

For comments, it's available along with the rest of the PR here: 
https://github.com/Azure/Azurite/pull/2228/files#diff-638244b70a10f551a0a6a66d691f8d88cf555fa0b9da1710e3d3bac119c05c39

### Performance

| Method         | Mean       | Error     | StdDev    |
|--------------- |-----------:|----------:|----------:|
| InMemory_Blob  | 1,860.0 us |  25.45 us |  21.25 us |
| InMemory_Queue | 1,562.3 us |  26.89 us |  25.15 us |
| InMemory_Table |   975.7 us |   8.65 us |   7.23 us |
| Disk_Blob      | 5,505.8 us |  91.37 us |  89.74 us |
| Disk_Queue     | 5,327.1 us | 104.47 us | 111.78 us |
| Disk_Table     |   963.6 us |  11.34 us |   9.47 us |

Unsurprisingly, blob and queue are much faster without writing extents to disk.


src/common/persistence/MemoryExtentStore.ts
```diff
@@ -0,0 +1,194 @@
...
+import IExtentStore, { IExtentChunk } from "./IExtentStore";
+import uuid = require("uuid");
+import { ReadableStreamBuffer } from 'stream-buffers'
+import multistream = require("multistream");
+
+export interface IMemoryExtentChunk extends IExtentChunk {
+  chunks: (Buffer | string)[]
+}
+
+export default class MemoryExtentStore implements IExtentStore {
```
joelverhagen, line 14: TODO: this needs debug logging like `FSExtentStore`

src/common/persistence/MemoryExtentStore.ts
```diff
@@ -0,0 +1,194 @@
...
+import { Readable } from "stream";
+
+export interface IMemoryExtentChunk extends IExtentChunk {
+  chunks: (Buffer | string)[]
+}
+
+export default class MemoryExtentStore implements IExtentStore {
+  private readonly metadataStore: IExtentMetadataStore;
+  private readonly logger: ILogger;
+  private readonly chunks: Map<string, IMemoryExtentChunk> = new Map<string, IMemoryExtentChunk>();
```
joelverhagen, line 17: It's unclear to me whether this should be and instance field or a static (process-wide) look-up. It seems to work fine instance level, I believe because the BlobServer/QueueServer/TableServer instances are effectively singletons per process (AFAIK)

blueww:
@joelverhagen 

Thanks for the contribution!
Would you please share the reason for this change? 
In which scenario we would like to have data cached in memory instead of disk?

And Azurite not only support saving data in disk, Azurite also allows save blob metadata in SQL, will the change also disable save data in SQL? (IF so, the parameter name/description might need change)
Besides that, have you tested all APIs of all services (B/T/Q) works with this change? If we plan to add this, we should have this kind of test environment covered in pipeline https://github.com/Azure/Azurite/blob/main/azure-pipelines.yml
And is there any limitation (like blob size) for the change? If so, we should mention it in the Readme.md file. 
And could you share how Azurite will handle it when out of memory in this case, what's the typical error?

joelverhagen:
> Would you please share the reason for this change?

This is intended for CI testing scenarios where Azurite is started at the beginning of the pipeline, some tests are run, and then the CI pipeline ends. 

Generally, I find myself using Azurite for cases where the I don't care much if the storage is lost or, if even annoys me to have to clean up leftover storage containers. Today, in my project NuGet/Insights (which I've used as a [general test for Azurite](https://github.com/Azure/Azurite/issues?q=is%3Aissue+is%3Aclosed+author%3Ajoelverhagen) as a side-goal) I have a step that cleans up containers, queues, and tables based on a test name convention. This is workable, and necessary for a test with real Azure Storage, but for CI pipelines it would be nice just to stop the node process and know all clean-up is done.
https://github.com/NuGet/Insights/blob/5d0b5ed62a71b10d93d77f5135b4876a1b1bf4dc/test/Worker.Logic.Test/TestSupport/TestDataTest.cs#L87-L114

Additionally, I've found that some of my tests run surprisingly slow and found that it was related to how queue messages are persisted in Azurite. A disk write is done per enqueue and dequeue which means there's a lot more IO than if LokiJS was used to persist message content, as Loki only writes periodically -- and this is noticeable with the current perf difference between Table and Queue write-then-read flows.

> In which scenario we would like to have data cached in memory instead of disk?

In short, CI runs where we start Azurite, run tests, then (optionally) stop Azurite.

> And Azurite not only support saving data in disk, Azurite also allows save blob metadata in SQL, will the change also disable save data in SQL? (IF so, the parameter name/description might need change)

It should still work. It does allow `--inMemoryPersistence` to work with SQL, where extent metadata is stored in SQL and the extents themselves are stored in memory. A weird combination for sure but the previous mode should work fine just by not specifying `--inMemoryPersistence`. From what I can tell, the `FSExtentStore` is used for SQL which also is a bit strange to me. It's essentially local extent storage and remote metadata storage. Anyways, happy to adjust the behavior however you'd like.

> Besides that, have you tested all APIs of all services (B/T/Q) works with this change? 

Yes, I did some test hacks to run all the tests I could with this change, but I'm not sure exactly how you'd like to get full coverage. Should we run all tests with and without the `--inMemoryPersistence` flag? That would double the number of tests but get very good coverage. Or we could run a subset with `--inMemoryPersistence`? Some representative subset? Happy to follow a suggestion here or I can give it a try on my own.

> Besides that, have you tested all APIs of all services (B/T/Q) works with this change? If we plan to add this, we should have this kind of test environment covered in pipeline https://github.com/Azure/Azurite/blob/main/azure-pipelines.yml

Great idea! I'll give that a try.  Maybe this addresses the previous point.

> And is there any limitation (like blob size) for the change? If so, we should mention it in the Readme.md file.

I'll take this as a TODO. It's certainly related to the amount of memory available to the node process.

> And could you share how Azurite will handle it when out of memory in this case, what's the typical error?

I'll take this as a TODO. I am not sure how node operates when it's low on memory. My guess is big allocs can fail and trigger GC and some severe memory shortage would cause the process to crash.

joelverhagen:
> > And could you share how Azurite will handle it when out of memory in this case, what's the typical error?

> I'll take this as a TODO. I am not sure how node operates when it's low on memory. My guess is big allocs can fail and trigger GC and some severe memory shortage would cause the process to crash.

I did testing. The approach is surprisingly scalable. When physical memory is consumed, virtual memory is provided by the OS. I was not able to get any "out of memory" errors. I included a screenshot in the design doc.

blueww:
@joelverhagen 

Thanks for the contribution!
We might will need some time to review/test the code change and the doc.

Will update you later.

joelverhagen:
Great, thanks @blueww! Let me know if there's any updates I should make or questions I can answer. I'm happy to jump on a Teams call too if you think that will help ease communication.

blueww:
Hi @joelverhagen,

I have done some sanity check with the PR:
1. I try to upload multiple 3GB blobs to Azurite run with "Azurite --inMemoryPersistence"
  - My machine is Windows 10 with 32GB memory with ~46GB virtual memory.
  - There are no OOM on my machine after I uploaded 16 * 3GB blob, when I upload the 17th 3GB blob, following error happens and Azurite exists. The behavior should changes between difference OS. 
  - It might not be good to eat almost all memory. Should we provide a way to limit the memory usage in this case (like 1/2 of current available memory + virtual memory).
 
```
<--- Last few GCs --->

[49684:00000229E11C69C0]   456817 ms: Mark-sweep (reduce) 195.0 (242.0) -> 195.0 (209.5) MB, 123.2 / 0.0 ms  (average mu = 0.033, current mu = 0.025) external memory pressure; GC in old space requested
[49684:00000229E11C69C0]   456893 ms: Mark-sweep (reduce) 195.0 (209.5) -> 195.0 (208.5) MB, 75.7 / 0.0 ms  (average mu = 0.019, current mu = 0.000) external memory pressure; GC in old space requested


<--- JS stacktrace --->

FATAL ERROR: Committing semi space failed. Allocation failed - JavaScript heap out of memory
 1: 00007FF6F8A4D51F node_api_throw_syntax_error+175743
 2: 00007FF6F89D2E66 v8::internal::wasm::WasmCode::safepoint_table_offset+59654
 3: 00007FF6F89D4B72 v8::internal::wasm::WasmCode::safepoint_table_offset+67090
 4: 00007FF6F947AAB4 v8::Isolate::ReportExternalAllocationLimitReached+116
 5: 00007FF6F9465E12 v8::Isolate::Exit+674
 6: 00007FF6F92E7C6C v8::internal::EmbedderStackStateScope::ExplicitScopeForTesting+124
 7: 00007FF6F92F20CC v8::internal::Heap::PageFlagsAreConsistent+3612
 8: 00007FF6F92E4707 v8::internal::Heap::CollectGarbage+2039
 9: 00007FF6F92FB0C3 v8::internal::HeapAllocator::AllocateRawWithLightRetrySlowPath+2099
10: 00007FF6F92FB96D v8::internal::HeapAllocator::AllocateRawWithRetryOrFailSlowPath+93
11: 00007FF6F930B1A3 v8::internal::Factory::NewFillerObject+851
12: 00007FF6F8FFC285 v8::internal::DateCache::Weekday+1349
13: 00007FF6F9518051 v8::internal::SetupIsolateDelegate::SetupHeap+558193
14: 00007FF6F94CEA38 v8::internal::SetupIsolateDelegate::SetupHeap+257624
15: 00007FF67977349A
```
2. After uploaded 11*3GB blobs , when Azurite exist (with ctrl+c), it use > 2 minutes to release all used memory and exist. Could we accelerate this? Or at least should mention it in doc to make everyone know about that. And not sure will any bad thing happen when kill Azurite before it's full exist.

4. The debug log (run with "--debug") looks disappear when run Azurite with "Azurite --inMemoryPersistence --debug debug.log" the debug log won't be generated in the current folder. 
  - Without "--inMemoryPersistence" the debug log will be generated in current folder. 
  - If specify full path of the debug log, the debug log will also be generated like "--debug c:\temp\debug.log"

5. To keep data alignment, we should not allow user to start Azurite with saving metadata to SQL (when environment variable "AZURITE_DB" is set) or specify the location parameter ("--location" or "-l").

joelverhagen:
Great feedback, thank you!

> It might not be good to eat almost all memory. Should we provide a way to limit the memory usage in this case (like 1/2 of current available memory + virtual memory).

I see `os.totalmem()` and `os.freemem()` available in the Node.js runtime. I don't see any way to probe the amount of virtual memory available. What do you think about adding a `--extentMemoryLimit <bytes>` option that allows the `MemoryExtentStore` to be limited in size? I don't see a way to limit the LokiJS persistence and now that I think about it any high memory usage of LokiJS would already be happening today.

If this option is left unspecified, it can dynamically limit the storage to say 50% of the currently available memory by querying `os.freemem()`. This would be lower than the allowance including virtual memory but perhaps that's fine in order to keep this simple.

> After uploaded 11*3GB blobs , when Azurite exist (with ctrl+c), it use > 2 minutes to release all used memory and exist. Could we accelerate this?

I faced the same. For my case, I believe it was due to the OS releasing memory even after the process was terminated. Because the process was terminated already and the memory releasing was still happening, I wasn't able to find a way to speed this up. Perhaps this will be less of concern if `--extentMemoryLimit` is implemented.

If you are seeing the release happen prior to the process completion, then perhaps we would improve the performance by concatenating the buffers provided to the memory extent store (in order to reduce the number of total buffers to release) but I feel that it's better to reduce copying memory for runtime performance and reducing the need for contiguous memory instead of optimizing for the speed to release memory at process end (which I generally imagine is a non-goal or at least low priority goal compared to other performance measures -- correct me if I am wrong).

> The debug log (run with "--debug") looks disappear when run Azurite with "Azurite --inMemoryPersistence --debug debug.log" the debug log won't be generated in the current folder.

I wasn't able to reproduce this problem. I ran `npm run azurite -- --inMemoryPersistence --debug debug.log` and the `debug.log` appeared just fine. 

> To keep data alignment, we should not allow user to start Azurite with saving metadata to SQL (when environment variable "AZURITE_DB" is set) or specify the location parameter ("--location" or "-l"), or at least omit the sql and location with a warning message.

Could you explain again what you mean here? Are you saying we shouldn't allow the combination of SQL metadata persistence and `--inMemoryStorage`? I'm totally fine with that, if I'm understanding you properly.

blueww:
> Great feedback, thank you!
> 
> > It might not be good to eat almost all memory. Should we provide a way to limit the memory usage in this case (like 1/2 of current available memory + virtual memory).
> 
> I see `os.totalmem()` and `os.freemem()` available in the Node.js runtime. I don't see any way to probe the amount of virtual memory available. What do you think about adding a `--extentMemoryLimit <bytes>` option that allows the `MemoryExtentStore` to be limited in size? I don't see a way to limit the LokiJS persistence and now that I think about it any high memory usage of LokiJS would already be happening today.
> 
> If this option is left unspecified, it can dynamically limit the storage to say 50% of the currently available memory by querying `os.freemem()`. This would be lower than the allowance including virtual memory but perhaps that's fine in order to keep this simple.
> 
I agree. And we should have good doc and API responds when memory is eat up. Like all write request will return failure with an error message to indicate the memory is eat up and user can increase the limitation by some way.

> > After uploaded 11*3GB blobs , when Azurite exist (with ctrl+c), it use > 2 minutes to release all used memory and exist. Could we accelerate this?
> 
> I faced the same. For my case, I believe it was due to the OS releasing memory even after the process was terminated. Because the process was terminated already and the memory releasing was still happening, I wasn't able to find a way to speed this up. Perhaps this will be less of concern if `--extentMemoryLimit` is implemented.
> 
> If you are seeing the release happen prior to the process completion, then perhaps we would improve the performance by concatenating the buffers provided to the memory extent store (in order to reduce the number of total buffers to release) but I feel that it's better to reduce copying memory for runtime performance and reducing the need for contiguous memory instead of optimizing for the speed to release memory at process end (which I generally imagine is a non-goal or at least low priority goal compared to other performance measures -- correct me if I am wrong).
> 
I think at least we should have good doc to indicate the behavior.
> > The debug log (run with "--debug") looks disappear when run Azurite with "Azurite --inMemoryPersistence --debug debug.log" the debug log won't be generated in the current folder.
> 
> I wasn't able to reproduce this problem. I ran `npm run azurite -- --inMemoryPersistence --debug debug.log` and the `debug.log` appeared just fine.
> 
Sorry, looks I can't repro it now.
> > To keep data alignment, we should not allow user to start Azurite with saving metadata to SQL (when environment variable "AZURITE_DB" is set) or specify the location parameter ("--location" or "-l"), or at least omit the sql and location with a warning message.
> 
> Could you explain again what you mean here? Are you saying we shouldn't allow the combination of SQL metadata persistence and `--inMemoryStorage`? I'm totally fine with that, if I'm understanding you properly.
1. The blob content/metadata will be saved in current folder by default, and when location parameter ("--location" or "-l") is specified, will be saved in the location folder, when environment variable "AZURITE_DB" is set, blob meta data (not content) will be saved in SQL.
So when specify "--inMemoryPersistence":
 -  location parameter will have not use, we should report warning for the parameter is not used and continue running. (or report error and stop running?)
 - When environment variable "AZURITE_DB" is set, will be blob meta data be saved to SQL, and blob content still in memory? IF so, in restart Azurite then blob metadata and blob content data will be not aligned. So we should block user from use both SQL and "--inMemoryPersistence" , or omit the SQL setting when user "--inMemoryPersistence" with a warning.
Feel free to let me know if you still have any concern/ question.

Besides that, our team will do more review/test with the PR, so more comments might will be added later.
One important thing is, the code change should not make any regression, which means when "--inMemoryPersistence" is not specified, Azurite behavior should not have any change from before in any case.

joelverhagen:
> I agree. And we should have good doc and API responds when memory is eat up. Like all write request will return failure with an error message to indicate the memory is eat up and user can increase the limitation by some way.

Done.

> I think at least we should have good doc to indicate the behavior.

Updated the README section that documents the option to mention this.

> location parameter will have no use, we should report warning for the parameter is not used and continue running. (or report error and stop running?)

Made it error out. `--location` + `--inMemoryPersistence` causes an error. `--extentMemoryLimit` without `--inMemoryPersistence` causes an error.

> We might can just omit the SQL setting with a warning (Or report error and stop running?).

`AZURITE_DB` + `--inMemoryPersistence` causes an error.

> One important thing is, the code change should not make any regression, which means when "--inMemoryPersistence" is not specified, Azurite behavior should not have any change from before .

This was my goal. The UTs still pass without any changes and in-memory persistence disabled and I did a test of a bunch of manual scenarios with the CLI and VS Code extension and this appears to be the case.

ChangeLog.md
```diff
@@ -4,6 +4,10 @@
 
 ## Upcoming Release
 
+General:
+
+- Add `--inMemoryPersistence` option and related configs to persist all data in-memory without disk persistence. (issue #2227)
```
blueww, line 9: Change log should also mention --extentMemoryLimit

README.md
```diff
@@ -14,12 +14,15 @@
...
-  - [Features & Key Changes in Azurite V3](#features--key-changes-in-azurite-v3)
+  - [Features \& Key Changes in Azurite V3](#features--key-changes-in-azurite-v3)
   - [Getting Started](#getting-started)
     - [GitHub](#github)
     - [NPM](#npm)
     - [Visual Studio Code Extension](#visual-studio-code-extension)
     - [DockerHub](#dockerhub)
+      - [Run Azurite V3 docker image](#run-azurite-v3-docker-image)
+      - [Run Azurite V3 docker image with customized persisted data location](#run-azurite-v3-docker-image-with-customized-persisted-data-location)
+      - [Customize all Azurite V3 supported parameters for docker image](#customize-all-azurite-v3-supported-parameters-for-docker-image)
```
blueww, line 25: To make the index not too long, we might can remove the 3 child sections under [DockerHub]

package.json
```diff
@@ -248,6 +248,19 @@
...
+            "default": false,
+            "description": "Disable persisting any data to disk. All data is stored in memory. If the Azurite (node) process is terminated or VS Code is closed, all data is lost."
+          },
+          "azurite.extentMemoryLimit": {
+            "type": [
+              "number",
+              "null"
+            ],
+            "default": null,
+            "description": "When using in-memory persistence, limit the total size of extents (blob and queue content) to a specific number of bytes. Defaults to 50% of total memory."
```
blueww, line 263: should be not only blob/queue, but also table.
As we might add more services in the future, might make the word more flexible like "(e.g. blob content)" should be better.

.vscode/launch.json
```diff
@@ -204,6 +217,34 @@
...
+      ],
+      "env": {
+        "AZURITE_ACCOUNTS": "",
+        "AZURE_TABLE_STORAGE": "",
+        "DATATABLES_ACCOUNT_NAME": "<name of your storage account>",
+        "DATATABLES_ACCOUNT_KEY" : "<account key for your storage account>",
+        "AZURE_DATATABLES_STORAGE_STRING": "https://<your account name>.table.core.windows.net",
+        "AZURE_DATATABLES_SAS": "?<sas query string>",
+        "NODE_TLS_REJECT_UNAUTHORIZED": "0",
+        "AZURITE_TEST_INMEMORYPERSISTENCE": "true"
```
blueww, line 243: This environment variable will only work in test.

README.md
```diff
@@ -430,6 +437,40 @@ Optional. When using FQDN instead of IP in request Uri host, by default Azurite
...
 
+### Use in-memory storage
+
+Optional. Disable persisting any data to disk and only store data in-memory. If the Azurite process is terminated, all
+data is lost. By default, LokiJS persists blob and queue metadata to disk and content to extent files. Table storage
+persists all data to disk. This behavior can be disabled using this option. This setting is rejected when the SQL based
+metadata implementation is enabled.
+
+```cmd
+--inMemoryPersistence
```
blueww, line 448: We should mention this parameter not work with some other parameters/environment variable.

README.md
```diff
@@ -430,6 +437,40 @@ Optional. When using FQDN instead of IP in request Uri host, by default Azurite
...
+```
+
+By default, the in-memory extent store (for blob and queue content) is limited to 50% of the total memory on the host
+machine. This is evaluated to using [`os.totalmem()`](https://nodejs.org/api/os.html#ostotalmem). This limit can be
+overridden using the `--extentMemoryLimit <bytes>` option. There is no restriction on the value specified for this
```
blueww, line 453: bytes maybe too small unit.

README.md
```diff
@@ -198,6 +203,8 @@ Following extension configurations are supported:
 - `azurite.oauth` OAuth oauthentication level. Candidate level values: `basic`.
 - `azurite.skipApiVersionCheck` Skip the request API version check, by default false.
 - `azurite.disableProductStyleUrl` Force parsing storage account name from request Uri path, instead of from request Uri host.
+- `azurite.inMemoryPersistence` Disable persisting any data to disk. If the Azurite process is terminated, all data is lost.
+- `azurite.memoryExtentLimit` When using in-memory persistence, limit the total size of extents (blob and queue content) to a specific number of bytes. Defaults to 50% of total memory.
```
blueww, line 207: Maybe we can mention which is not limited by this parameter.

package.json
blueww: I get the limit is only for content, but not for meta data.

src/table/TableServer.ts
```diff
@@ -73,6 +74,20 @@ export default class TableServer extends ServerBase {
     this.accountDataStore = accountDataStore;
   }
 
+  public async clean(): Promise<void> {
```
blueww, line 77: This fix a bug: call table.clean in VS code extention will not work before the fix.

src/table/main.ts
```diff
@@ -78,7 +79,7 @@ async function main() {
     })
     .once("SIGINT", () => {
       console.log(beforeCloseMessage);
-      server.clean().then(() => {
+      server.close().then(() => {
```
blueww, line 82: This fix another table but, control+c will clean table instead of close table.

tests/blob/blockblob.highlevel.test.ts
```diff
@@ -78,14 +78,14 @@ describe("BlockBlobHighlevel", () => {
     }
     tempFileLarge = await createRandomLocalFile(
       tempFolderPath,
-      257,
-      1024 * 1024
+      8224,
+      1024 * 32
```
blueww, line 82: The blob count and blob size change, although the blob size not change.
Need look into if this will lost some test coverage.

tests/blob/blockblob.highlevel.test.ts
```diff
@@ -306,7 +306,7 @@ describe("BlockBlobHighlevel", () => {
     const aborter = new AbortController();
     try {
       await blockBlobClient.downloadToBuffer(buf, 0, undefined, {
-        blockSize: 1 * 1024,
+        blockSize: 32 * 1024, // if too small, the test is very slow
```
blueww, line 309: Better to revert that for test coverage.

tests/exe.test.ts
```diff
@@ -53,6 +54,14 @@ configLogger(false);
...
 const testLocalAzuriteInstance = true;
 
+const binaryPath = ".\\release\\azurite.exe"
+
+function throwOnMissingBinary() {
+  if (!existsSync(binaryPath)) {
+    throw new Error("The Windows binary does not exist. You must build it first using 'npm run build:exe'.")
+  }
+}
+
```
blueww, line 64: This fix a test hang when azurite.exe doesn't exist.

tests/queue/oauth.test.ts
```diff
@@ -20,38 +20,26 @@ describe("Queue OAuth Basic", () => {
...
   const baseURL = `https://${host}:${port}/devstoreaccount1`;
 
   before(async () => {
-    server = new Server(config);
+    server = new QueueTestServerFactory().createServer({
+      metadataDBPath: metadataDbPath,
+      extentDBPath: extentDbPath,
+      persistencePathArray: DEFAULT_QUEUE_PERSISTENCE_ARRAY,
+      https: true,
+      oauth: "basic"
```
blueww, line 41: Wei need look at cert/key value pass through.

tests/queue/utils/QueueTestServerFactory.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export default class QueueTestServerFactory {
+  public createServer(params: IQueueTestServerFactoryParams): QueueServer {
+    const inMemoryPersistence = process.env.AZURITE_TEST_INMEMORYPERSISTENCE !== undefined;
+
+    const port = 11001;
+    const host = "127.0.0.1";
+
+    const cert = params.https ? "tests/server.cert" : undefined;
+    const key = params.https ? "tests/server.key" : undefined;
```
blueww, line 25: This is just blob test do.

ChangeLog.md
joelverhagen: Fixed.

README.md
joelverhagen: Fixed.

README.md
joelverhagen: Fixed.

tests/blob/blockblob.highlevel.test.ts
joelverhagen: Fixed.

src/queue/handlers/MessageIdHandler.ts
```diff
@@ -5,7 +5,7 @@ import Context from "../generated/Context";
 import IMessageIdHandler from "../generated/handlers/IMessageIdHandler";
 import { MessageUpdateProperties } from "../persistence/IQueueMetadataStore";
 import {
-  DEFUALT_UPDATE_VISIBILITYTIMEOUT,
+  DEFAULT_UPDATE_VISIBILITYTIMEOUT,
```
joelverhagen, line 8: This was misspelled.

src/queue/handlers/MessagesHandler.ts
```diff
@@ -7,8 +7,8 @@ import Context from "../generated/Context";
 import IMessagesHandler from "../generated/handlers/IMessagesHandler";
 import { MessageModel } from "../persistence/IQueueMetadataStore";
 import {
-  DEFUALT_DEQUEUE_VISIBILITYTIMEOUT,
-  DEFUALT_MESSAGETTL,
+  DEFAULT_DEQUEUE_VISIBILITYTIMEOUT,
+  DEFAULT_MESSAGETTL,
```
joelverhagen, line 11: This was misspelled.

README.md
joelverhagen: Change it to megabytes.

README.md
joelverhagen: Done.

joelverhagen:
Hey @blueww! I believe I have addressed all of your comments. Please let me know if more is needed 😃

src/common/ConfigurationBase.ts
```diff
@@ -1,12 +1,30 @@
...
   PFX
 }
 
+export function setExtentMemoryLimit(env: IBlobEnvironment | IQueueEnvironment) {
+  if (env.inMemoryPersistence()) {
+    let mb = env.extentMemoryLimit() ?? SharedChunkStore.sizeLimit();
+    if (mb && mb >= 0) {
+      const bytes = Math.round(mb * 1024 * 1024);
+      const totalPct = Math.round(100 * bytes / totalmem())
+      console.log(`In-memory extent storage is enabled with a limit of ${mb.toFixed(2)} MB (${bytes} bytes, ${totalPct}% of total memory).`);
```
blueww, line 20: The log should also be print to Debug log.

src/common/ConfigurationBase.ts
```diff
@@ -1,12 +1,30 @@
...
+      SharedChunkStore.setSizeLimit(bytes);
+    } else {
+      console.log(`In-memory extent storage is enabled with no limit on memory used.`);
```
blueww, line 23: The log should also be print to Debug log.

src/common/Environment.ts
```diff
@@ -155,6 +166,24 @@ export default class Environment implements IEnvironment {
...
+      return true;
+    } else {
+      if (this.extentMemoryLimit() !== undefined) {
+        throw new RangeError(`The --extentMemoryLimit option is only supported when the --inMemoryPersistence option is set.`)
+      }
+    }
+    return false;
+  }
+
+  public extentMemoryLimit(): number | undefined {
```
blueww, line 183: Should check the number >= 0.

src/common/VSCServerManagerClosedState.ts
```diff
@@ -9,6 +11,18 @@ export default class VSCServerManagerClosedState
   ): Promise<IVSCServerManagerState> {
     await serverManager.createImpl();
     await serverManager.startImpl();
+
+    const limit = new VSCEnvironment().extentMemoryLimit()
+    if (limit) {
+      if (limit >= 0) {
+        SharedChunkStore.setSizeLimit(Math.round(limit * 1024 * 1024))
+      } else {
+        SharedChunkStore.setSizeLimit()
```
blueww, line 20: What will happen if user set a negative value? We might should remind user the value is invalid.

tests/blob/blockblob.highlevel.test.ts
blueww: Discuss with Xiaoning, this should be OK.

src/common/VSCServerManagerClosedState.ts
joelverhagen: We can choose the behavior. It felt to me like a reasonable way to specify "no limit" by passing a negative value. `-1` is a common marker value in other contexts. It's also simpler to implement since we don't have to think about how to best show the error message in the VS Code options UI.

joelverhagen: For now, I will document the behavior.

joelverhagen: It looks like negative numbers don't work well on the CLI because `-1` get interpreted as an option called "-1" not a negative option value. I'll make it reject the negative number.

src/common/ConfigurationBase.ts
joelverhagen: Done

src/common/ConfigurationBase.ts
joelverhagen: Done

src/common/VSCServerManagerClosedState.ts
joelverhagen: Done

blueww:
@joelverhagen 

I have tested repeated upload and delete blobs with Azurite run with "--inMemoryPersistence". But the test result is not expected, it looks the deleted blob won't be clear up from memory. Would you please help to look?

#### Test Scenario:
Upload 100MB * 10 blobs, then delete all 10 blobs. Then upload 100MB* 10 blobs again, and delete them again. ...
I tried to repeat 1000 times, but failed after ~15 times.

##### Test Result:
After upload/delete 100MB*10 blobs for ~15 times, will meet following error :

```
Set-AzStorageBlobContent : Cannot add an extent chunk to the in-memory store. Size limit of 17030338560 bytes will be exceeded HTTP Status Code: 409 - HTTP Error Message: Cannot add an extent chunk to the in-memory store. Size limit of 17030338560 bytes will be exceeded
ErrorCode: MemoryExtentStoreAtSizeLimit
ErrorMessage: Cannot add an extent chunk to the in-memory store. Size limit of 17030338560 bytes will be exceeded
RequestId:53834723-e10f-48f9-aa9d-35330109a991
Time:2023-11-13T08:33:50.957Z
At line:6 char:9
+         Set-AzStorageBlobContent -Container $containerName -Blob "tes ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : CloseError: (:) [Set-AzStorageBlobContent], StorageException
    + FullyQualifiedErrorId : StorageException,Microsoft.WindowsAzure.Commands.Storage.Blob.SetAzureBlobContentCommand
```

The memory usage chart (I wait > 15 minutes after error happen) :
image


For you reference, my test script:
I run it from Powershell, need first install [Az.Storage module](https://www.powershellgallery.com/packages/Az.Storage/5.10.1).
```powershell
$ctx = New-AzStorageContext -ConnectionString "AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;DefaultEndpointsProtocol=http;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;QueueEndpoint=http://127.0.0.1:10001/devstoreaccount1;TableEndpoint=http://127.0.0.1:10002/devstoreaccount1;"

$containerName = "test"
New-AzStorageContainer -Name $containerName -Context $ctx

for ($i =0 ; $i -le 1000; $i++ )
{
    Echo "create Blobs"
    for ($j=0 ; $j -le 10; $j++)
    {
        Set-AzStorageBlobContent -Container $containerName -Blob "testblob_$($i)_$($j)" -File C:\temp\testfile_102400K_0 -Context $ctx -Force | ft
    }
    echo "Delete blobs"
    for ($j=0 ; $j -le 10; $j++)
    {
        Remove-AzStorageBlob -Container $containerName -Blob "testblob_$($i)_$($j)"  -Context $ctx -PassThru -Force | ft
    }
}
```

joelverhagen:
@blueww - the default blob GC is 10 minutes (https://github.com/Azure/Azurite/blob/49d1065952306c25ce76f84e041cd1acd2e951ac/src/blob/utils/constants.ts#L21). From your memory chart, I see growth in memory only 3-4 minutes after the process start. It is expected to face this error if you have met the limit before the `BlobGCManager` runs. If you were to attempt the upload after 10 minutes from process start, you should see success. I see a spike in CPU right around the 10 minute mark so I'm guessing the blob GC actually happened but your uploads did not continue after you faced an error (~3-4 minute mark).

Regarding the memory consumption, I wasn't able to reproduce this behavior. I changed the GC period to 5 minutes to ease testing and set an extent limit of 5 GB.
image
(I am not sure what tool you used, but it looks like a visualization of system perf counters)

A debug log will shed light on this. You'll want to look for a line like this:
```
info: BlobGCManager:markSweep() Deleted unreferenced 650 extents, after excluding active write extents.
```

It's possible that Node isn't releasing the memory even after the blob GC occurs. I am not an expert in Node.js GC but after a bit of reading it seems like it could be related to the default `max-old-space-size` value selected on your machine. It seems Node.js won't perform some of the more expensive GC steps if the current memory consumption is lower than this threshold which may explain the behavior you're seeing. The part of the code that is impacted by this PR is the integration with the blob and the queue GC. If you see the `markSweep` log line similar to the above, then it's working as I designed it. That being said, if you provide a debug log it might help me understand what you're seeing.

A way you could run a similar test for the existing disk-based persistence. I tried with a ramdisk on Ubuntu WSL (steps: https://askubuntu.com/a/453755) and similarly encountered failures (HTTP 500 not 409 of course) at my 5 GB ramdisk limit. However in that case the blob GC didn't find any extents at all. I tried many versions of Azurite (historical tags like `v3.10.0` with the same condition:
```
info: BlobGCManager:markSweep() Got 0 extents.
```
So an alternate explanation is that there is an existing bug in the blob GC manager that you encountered. But we need a debug log to know.

blueww:
@joelverhagen 

Thanks for the investigation!

I use the tool [Perfmon ](https://techcommunity.microsoft.com/t5/ask-the-performance-team/windows-performance-monitor-overview/ba-p/375481)to show the chart of memory/cpu.

I have tested following 3 scenario, and this time looks GC works in all scenarios:
#### Scenario 1
1. upload 100*100MB blobs, them delete them
2. Wait for 10 minutes
3. repeated step 1, step 2.

image

#### Scenario 2
1. Upload 200*100MB blobs (after upload ~150 blobs will get OOM error) and delete them, 
2. wait 10 minutes 
3. repeated step 1, step 2. 
image

#### Scenario 3
1. Upload 10*100MB blobs  and delete them, 
2. Repeat step 1 for 20 times (after repeat 15 times will get OOM error)
3. wait 1000 seconds
4. repeated step 1~3.
We can see the "Private Bytes" won't drop immediately after GC, this is different from other 2 scenarios. But it will drop in the begin of next round blob upload.  
image

joelverhagen:
> Besides that, to avoid customer issues like why deleted blob content still occupy memory, do you think we should mention how GC works in Readme.md section "Use in-memory storage"?

Sure, I'll do that.

> We can see the "Private Bytes" won't drop immediately after GC, this is different from other 2 scenarios. But it will drop in the begin of next round blob upload.

Generally I've seen this behavior in several GC'd runtimes. In my head it's basically "GC sometimes does the hard work only when memory pressure is faced" (i.e. lazily) -- which is when the next round of blobs come in for this case. But this I my gut check and not specific expertise on the Node GC.

blueww:
@joelverhagen 

Thanks for the quick update!
I will do some final review, and should can merge this PR soon.

blueww (approved):

edwin-huber:
Fixes #1331
Fixes #1332

Issue #1967, opened 2023-05-12
500 Errors - Error: EMFILE: too many open files


blperf:
### Which service(blob, file, queue, table) does this issue concern?
Blob

### Which version of the Azurite was used?
3.20.1

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
Visual Studio 

### What's the Node.js version?
n/a

### What problem was encountered?
error: ErrorMiddleware: ErrorName=Error ErrorMessage=EMFILE: too many open files, open 'c:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\Extensions\Microsoft\Azure Storage Emulator\__blobstorage__\74924663-2a08-4213-8c52-74637ab9dd77' ErrorStack="Error: EMFILE: too many open files, open 'c:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\Common7\\IDE\\Extensions\\Microsoft\\Azure Storage Emulator\\__blobstorage__\\74924663-2a08-4213-8c52-74637ab9dd77'"

### Steps to reproduce the issue?
From debug log:
2023-04-19T00:09:59.105Z 831912d4-1317-445d-b45d-e9f6e62e8566 error: ErrorMiddleware: Received an error, fill error information to HTTP response
2023-04-19T00:09:59.105Z 831912d4-1317-445d-b45d-e9f6e62e8566 error: ErrorMiddleware: ErrorName=Error ErrorMessage=EMFILE: too many open files, open 'c:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\Extensions\Microsoft\Azure Storage Emulator\__blobstorage__\74924663-2a08-4213-8c52-74637ab9dd77' ErrorStack="Error: EMFILE: too many open files, open 'c:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\Common7\\IDE\\Extensions\\Microsoft\\Azure Storage Emulator\\__blobstorage__\\74924663-2a08-4213-8c52-74637ab9dd77'"
2023-04-19T00:09:59.105Z 831912d4-1317-445d-b45d-e9f6e62e8566 error: ErrorMiddleware: Set HTTP code: 500

### Have you found a mitigation/solution?
Killing and restarting Azurite resolves the issue immediately, but I'm looking for a better solution to avoid this in the first place. Is there anything that can be done to prevent this? SAS durations etc? Any way to monitor how many files it has open to see when the limit is being approached? 

[...]

PR #2267, opened 2023-10-29
Adding initial fix for bug issue #1967

blperf:

### PR Branch Destination

### Always Add Test Cases

### Add Change Log

### Development Guideline

### DIFF
--- a/src/blob/generated/ExpressMiddlewareFactory.ts
+++ b/src/blob/generated/ExpressMiddlewareFactory.ts
@@ -92,13 +92,15 @@ export default class ExpressMiddlewareFactory extends MiddlewareFactory {
       handlers,
       this.logger
     );
-    return (req: Request, res: Response, next: NextFunction) => {
+      return (req: Request, res: Response, next: NextFunction) => {
       const request = new ExpressRequestAdapter(req);
       const response = new ExpressResponseAdapter(res);
+      let newContext = new Context(res.locals, this.contextPath, request, response);
       handlerMiddlewareFactory.createHandlerMiddleware()(
-        new Context(res.locals, this.contextPath, request, response),
+        newContext,
         next
       );
+      res.on("close", () => handlers.blobHandler.cleanUpBlob(newContext));
     };
   }
 
--- a/src/blob/generated/handlers/IBlobHandler.ts
+++ b/src/blob/generated/handlers/IBlobHandler.ts
@@ -14,6 +14,7 @@ import Context from "../Context";
 
 export default interface IBlobHandler {
   download(options: Models.BlobDownloadOptionalParams, context: Context): Promise<Models.BlobDownloadResponse>;
+  cleanUpBlob(context: Context): Promise<void>;
   getProperties(options: Models.BlobGetPropertiesOptionalParams, context: Context): Promise<Models.BlobGetPropertiesResponse>;
   delete(options: Models.BlobDeleteMethodOptionalParams, context: Context): Promise<Models.BlobDeleteResponse>;
   undelete(options: Models.BlobUndeleteOptionalParams, context: Context): Promise<Models.BlobUndeleteResponse>;
--- a/src/blob/handlers/BlobHandler.ts
+++ b/src/blob/handlers/BlobHandler.ts
@@ -91,6 +91,15 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
     }
   }
 
+    /**
+   * Clean up file handles for blob to prevent leak.
+   *
+   * @memberof BlobHandler
+   */
+    public async cleanUpBlob(context: Context): Promise<void> {
+        this.extentStore.cleanStreams(context.contextId);
+    }
+
   /**
    * Get blob properties.
    *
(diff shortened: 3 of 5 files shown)


src/blob/generated/ExpressMiddlewareFactory.ts
```diff
@@ -92,13 +92,15 @@ export default class ExpressMiddlewareFactory extends MiddlewareFactory {
...
       this.logger
     );
-    return (req: Request, res: Response, next: NextFunction) => {
+      return (req: Request, res: Response, next: NextFunction) => {
       const request = new ExpressRequestAdapter(req);
       const response = new ExpressResponseAdapter(res);
+      let newContext = new Context(res.locals, this.contextPath, request, response);
       handlerMiddlewareFactory.createHandlerMiddleware()(
-        new Context(res.locals, this.contextPath, request, response),
+        newContext,
```
EmmaZhu, line 100: Any reason change this?

EmmaZhu:
Hi @blperf , 

Really appreciate your contribution.

I tried to test the change with code like:
```
    await blockBlobClient.uploadFile("D:\\a\\ALargeFile");
    let result = await blobClient.download(0);
    await bodyToString(result, content.length);
    console.log("First download completed");

    result = await blobClient.download(0);
    await bodyToString(result, content.length);
    console.log("Second download completed");
```

The `ALargeFile` is about 27GB, still meet the error: 
```
2023-10-31T06:48:34.816Z 333c4e83-eb5b-4efb-bbe4-2ec68b0819f0 error: ErrorMiddleware: Received an error, fill error information to HTTP response
2023-10-31T06:48:34.816Z 333c4e83-eb5b-4efb-bbe4-2ec68b0819f0 error: ErrorMiddleware: ErrorName=Error ErrorMessage=EMFILE: too many open files, open 'D:\repo\azurite\__test_blob_extent__\d9d3aca7-62c7-470f-a6ab-59a8500936be' ErrorStack="Error: EMFILE: too many open files, open 'D:\\repo\\azurite\\__test_blob_extent__\\d9d3aca7-62c7-470f-a6ab-59a8500936be'"
2023-10-31T06:48:34.817Z 333c4e83-eb5b-4efb-bbe4-2ec68b0819f0 error: ErrorMiddleware: Set HTTP code: 500
```

When the blob size is large enough, it will still meet the error. 

From what I investigated, to fully fix the issue, we may need to change to "open stream, read it and close handle" when sending response body.  Do you have any thoughts?

Thanks
Emma

blperf:
Hi @EmmaZhu,

I see what you mean, I'll summarize to the best of my current understanding what is going on here. Before the fix in this PR, even smaller files when accessed too many times (using blobClient.download for example) would eventually result in enough open file descriptors to hit the OS limit. Because they were never closed, this would inevitably occur at some point with enough requests regardless of the file size, but happened sooner for larger files because more extents are being read per request. 

A less likely but still valid scenario you have brought up is, what happens if one single file is large enough that the number extents that needs to be read during a single blobClient.download request creates more open file descriptors than the OS limit?
In this situation, res.on("close") would not be hit before the number of extents being read would exceed the OS limit and the same error would be thrown. As you mentioned, this is for very large files.

I think you've already correctly pointed out the only way to avoid this scenario, which would be to close the file descriptor immediately after reading each extent before sending it back to the client. Do you have any links you could share from your research on how this is typically implemented?

Do you have any other thoughts on the best way to implement this? I'm wondering if there's some way to do this without changing the return type of readExtent, or if the return type would need to be changed. I'm not sure if it's possible to return a NodeJS.ReadableStream without keeping a file descriptor open, maybe it is somehow by reading from a value buffered in memory?

If you already have some other code for this, please feel free to share a PR link and I would be happy to abandon this one if another better fix is available. I'm certainly willing to verify another fix resolves this issue on my side as well. If not I can potentially try to help rewrite this PR to something along the lines of the strategy mentioned above.

EmmaZhu:
@blperf  Thanks a lot for such detailed summarize. Yes, that's what I mean.

For the solution, I'm think of implementing a readableStream to only open file when it needs to read content and close the file right after the reading is completed to replace stream returned here: 

https://github.com/Azure/Azurite/blob/main/src/common/persistence/FSExtentStore.ts#L334
https://github.com/Azure/Azurite/blob/main/src/common/persistence/FSExtentStore.ts#L343

Another way to fix it can be: having a file path array instead of a readableStream array, and only create a readable stream to a specific extent file when needs to read its content.

There might be other ways, as long as it only access the file when really necessary. 

It may require more changes to completely fix the issue.  We'd really appreciate if you are interested and available in fixing it. If not, I'll work on it in several days. 

Thanks
Emma

blperf:
Hi @EmmaZhu,

Happy new year, I hope all is well. Have you by any chance had time to take a look at this? I unfortunately haven't had time to try implementing your proposed solution, but I can confirm I have also encountered some other errors eventually using the fix in this PR. It takes much longer before the issues arise, but unfortunately it seems like this PR wasn't a complete fix. Perhaps you can start a new PR with a draft of the implementation you had in mind? I'm happy to do my best to contribute to the PR and help out where I can and can also test things out on my end to help ensure the original bug is resolved.

PR #2302, opened 2023-11-22
Fix HTTP header parsing in SubmitBatch request

kou:
If a SubmitBatch request has a HTTP header that has a HTTP header
delimiter (":") in its value, the SubmitBatch request is failed as
"400 One of the request inputs is not valid."

For example, if `user-agent` header is `azsdk-cpp-storage-blobs/12.10.0-beta.1 (Darwin 23.1.0 arm64 Darwin Kernel Version 23.1.0: Mon Oct 9 21:28:12 PDT 2023; root:xnu-10002.41.9~6/RELEASE_ARM64_T8103)`, all `SubmitBatch()` requests are failed.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,6 +13,7 @@ General:
 Blob:
 
 - Fixed issue of not requiring SAS permission for some specific operations. (issue #2299)
+- Fix HTTP header parsing of `SubmitBatch()`. If a HTTP header has HTTP header delimiter (`:`) in its value, `SubmitBatch()` returns "400 One of the request inputs is not valid". For example, if `user-agent` header is `azsdk-cpp-storage-blobs/12.10.0-beta.1 (Darwin 23.1.0 arm64 Darwin Kernel Version 23.1.0: Mon Oct  9 21:28:12 PDT 2023; root:xnu-10002.41.9~6/RELEASE_ARM64_T8103)`, all `SubmitBatch()` requests are failed.
 
 Table:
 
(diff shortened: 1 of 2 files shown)


tests/blob/apis/blobbatch.test.ts
```diff
@@ -32,7 +32,10 @@ describe("Blob batch API", () => {
       {
         retryOptions: { maxTries: 1 },
         // Make sure socket is closed once the operation is done.
-        keepAliveOptions: { enable: false }
+        keepAliveOptions: { enable: false },
+        // Make sure HTTP headers are parsed properly with many HTTP
+        // header delimiters.
+        userAgentOptions: { userAgentPrefix: "include:HTTP:header:delimiter " }
```
kou, line 38: Hmm... It seems that this doesn't change `user-agent` header in `SubmitBatch` request...

Could someone advice me how to write a test for this case? I want to add `header-name: value:value` header to `SubmitBatch` request...

blueww:
@EmmaZhu 

Would you please help to look at the fix for blob batch code?

tests/blob/apis/blobbatch.test.ts
EmmaZhu: The JS SDK doesn't support to set headers in sub requests.  
Seems you encountered the issue when you are using CPP sdk?
It should be fine if you can share your repro project and I can manually test it.

EmmaZhu:
I've manually tested the fixed. It works well with a header like `testheader: testvalue:test:value` in sub requests.  Will approve the PR.
@blueww , do you have other concerns?

felipecrv:
I started hitting this in my unit tests as well now.

felipecrv (approved):

PR #2316, opened 2023-12-07
Bugfix: 1385: Authentication fails if both date and x-ms-date headers are present

BillBooks:
For queues and blobs, when forming the shared key token, the Date header should be treated as the empty string when the x-ms-date header is present.

For tables, the value of the x-ms-date header should be used as the value of the Date header if the x-ms-date header is present.

This behavior is documented at
https://learn.microsoft.com/en-us/rest/api/storageservices/authorize-with-shared-key


blueww (changes requested):
The change generally looks good to me.
It's better to also add test cases for blob/Queue servicce.

BillBooks:
I'll look at adding some tests

jainakanksha-msft (approved):

PR #2330, opened 2023-12-26
Fix issue of copy succeeded without 'r' permission in SAS token credential

EmmaZhu:

### PR Branch Destination

### Always Add Test Cases

### Add Change Log

### Development Guideline

### DIFF
--- a/src/blob/handlers/BlobHandler.ts
+++ b/src/blob/handlers/BlobHandler.ts
@@ -122,39 +122,39 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
 
     const response: Models.BlobGetPropertiesResponse = againstMetadata
       ? {
-          statusCode: 200,
-          metadata: res.metadata,
-          eTag: res.properties.etag,
-          requestId: context.contextId,
-          version: BLOB_API_VERSION,
-          date: context.startTime,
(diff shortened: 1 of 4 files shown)


src/blob/handlers/BlobHandler.ts
```diff
@@ -664,7 +662,8 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
       throw StorageErrorFactory.getBlobNotFound(context.contextId!);
     }
 
-    if (sourceAccount !== blobCtx.account) {
+    const sig = url.searchParams.get("sig");
+    if ((sourceAccount !== blobCtx.account) || (sig !== null)) {
```
blueww, line 666: Just to confirm, this is the only product by propose. 
There are so many coding format changes, and not easy to get all real product code change.

blueww (approved):

src/blob/handlers/BlobHandler.ts
EmmaZhu: Yes, this is the only place I changed for product code. Other are all changed by prettier tool .

arnabfromjec05:
@EmmaZhu 
which build version will this bug fix be released?

PR #2377, opened 2024-03-15
Update azure-pipelines.yml for Azure Pipelines

EmmaZhu:


azure-pipelines.yml
```diff
@@ -572,7 +572,7 @@ jobs:
         # node_10_x:
         #   node_version: 10.x
         node_14_x:
-          node_version: 14.x
+          node_version: 20.x
```
blueww, line 575: The above line also need change to 20 to be aligned with node version it really use.

Issue #2053, opened 2023-07-13
unexptected TCP RST from azurite side


Slach:
### Which service(blob, file, queue, table) does this issue concern?
blob
### Which version of the Azurite was used?
latest docker image
### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
I created reproducible docker-compose
https://gist.github.com/Slach/043c85f759d4422a75fceab6dccb1f8d

### What problem was encountered?
TCP RST after make client requests
![image](https://github.com/Azure/Azurite/assets/105560/9bc9ecb2-f9b3-400d-898e-2017708c0118)
![image](https://github.com/Azure/Azurite/assets/105560/3b3a4876-6a26-433f-b120-c4e7b4ced836)

### Steps to reproduce the issue?
```
git clone https://gist.github.com/Slach/043c85f759d4422a75fceab6dccb1f8d
./reproduce.sh
```

### Have you found a mitigation/solution?
No

also look details https://github.com/ClickHouse/ClickHouse/issues/52088

blueww:
@Slach 
Thanks for raising this issue!

Would you please share the Azurite debug log for this issue? 
And please point out in which step of reproduce.sh does this issue repro?

Besides that, just to confirm this kind of issue not happen on product Azure?

parshimers:
I've experienced this issue as well in a different context. The debug log shows nothing, because the connection is immediately terminated by the RST. It happens immediately after the client connects. I can't reproduce it on MacOS but it happens easily on Linux.

blueww:
@parshimers 

Do you use the same Azurite docker image version on linux and maxOS? (Docker image with version latest might be cached in local. IF so you might can clearup the local cached Azurite latest docker image, and try again to make sure you use latest docker image.)

As I also can't repro this issue locally, it will be very difficult for us to investigate. 
As you can't repro on MacOS but can repro on linux (I assume you use same docker image), it looks not an Azurite issue, but some issue related with your OS environment. 

Besides that, does the issue happen randomly, how frequently? If you add some retry logic on storage client side, can it workaround this issue?

PR #2443, opened 2024-08-01
add blobKeepAliveTimeout, queueKeepAliveTimeout, tableKeepAliveTimeout

Slach:
fix https://github.com/Azure/Azurite/issues/2053 
workaround for https://github.com/ClickHouse/ClickHouse/issues/60447

we use azurite for Azure BLOB storage in our integration tests
Azure CPP Sdk (or it usage on clickhouse-server side) don't respect `Keep-Alive: 5` header which sent from azurite side, and continue to make request to already closed connection after 5 seconds, this header sent from standard nodejs http library

We are going to fix this behavior in the next version of `clickhouse-server`
but currently to allow pass e2e tests, we would like to propose adding cli option and config settings to allow increase value for Keep-Alive header

we tested these changes and it actually works, hope your CI/CD pipelines also will pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,10 +186,13 @@ Following extension configurations are supported:
 
 - `azurite.blobHost` Blob service listening endpoint, by default 127.0.0.1
 - `azurite.blobPort` Blob service listening port, by default 10000
+- `azurite.blobKeepAliveTimeout` Blob service keep alive timeout, by default 5
 - `azurite.queueHost` Queue service listening endpoint, by default 127.0.0.1
 - `azurite.queuePort` Queue service listening port, by default 10001
+- `azurite.queueKeepAliveTimeout` Queue service keep alive timeout, by default 5
 - `azurite.tableHost` Table service listening endpoint, by default 127.0.0.1
 - `azurite.tablePort` Table service listening port, by default 10002
+- `azurite.tableKeepAliveTimeout` Queue service keep alive timeout, by default 5
 - `azurite.location` Workspace location folder path (can be relative or absolute). By default, in the VS Code extension, the currently opened folder is used. If launched from the command line, the current process working directory is the default. Relative paths are resolved relative to the default folder.
 - `azurite.silent` Silent mode to disable access log in Visual Studio channel, by default false
 - `azurite.debug` Output debug log into Azurite channel, by default false
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -15,6 +15,15 @@ Blob:
 - Fixed issue of download a blob range without header x-ms-range-get-content-md5, should not return content-md5. (issue #2409)
 - Fixed issue of list container without include=metadata should not clear container metadata on server. (issue #2416)
 - Supported x-ms-copy-source-tag-option in copy blob from Uri. (issue #2398)
+- Added blobKeepAliveTimeout option (issue #2053)
+
+Table:
+
+- Added tableKeepAliveTimeout option (issue #2053)
+
+Queue:
+
+- Added queueKeepAliveTimeout option (issue #2053)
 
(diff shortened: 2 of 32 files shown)


Slach:

blueww:
@Slach 

Thanks for the contribution!
Would you please rebase the code on the latest main to make PR validation can run?

And could you give more details on why the change can fix the issue?

Slach:
rebased

Slach:
PR description updated

blueww:
@Slach 

Besides test class "FSExtentStore" failure on  BlobTest_Win_Node* is know issue which is in fixing, other test failures are new. 
Please help to fix the test failures.

Most of the new failure looks like :
```
  1) Blob Cors requests test
       Request Match rule exists for exact origin @loki @sql:
     FetchError: request to http://127.0.0.1:11000/devstoreaccount1?restype=service&comp=properties failed, reason: read ECONNRESET
      at ClientRequest.<anonymous> (node_modules/node-fetch/lib/index.js:1491:11)
      at ClientRequest.emit (node:events:519:28)
      at ClientRequest.emit (node:domain:488:12)
      at Socket.socketErrorListener (node:_http_client:500:9)
      at Socket.emit (node:events:519:28)
      at Socket.emit (node:domain:488:12)
      at emitErrorNT (node:internal/streams/destroy:169:8)
      at emitErrorCloseNT (node:internal/streams/destroy:128:3)
      at processTicksAndRejections (node:internal/process/task_queues:82:21)
```

README.md
```diff
@@ -186,10 +186,13 @@ Following extension configurations are supported:
 
 - `azurite.blobHost` Blob service listening endpoint, by default 127.0.0.1
 - `azurite.blobPort` Blob service listening port, by default 10000
+- `azurite.blobKeepAliveTimeout` Blob service keep alive timeout, by default 5
```
blueww, line 189: What's the value before this PR?
Could we use the value before this PR is merged to avoid regression?

Slach: it was 5 by default

Slach:
sorry
i don't understand how to interpet stacktrace in https://github.com/Azure/Azurite/pull/2443#issuecomment-2264793405

how to match which test exactly doesn't work? And how to run only this failed test?

i tried to run  `npm run test:blob`  locally on node 20

and got lot of failures can't figure out how to debug it

blueww:
@Slach 

If you click the "Details" of the failed jobs, you should can see the failure log, and find the failed test class/case name.
image

You might can first focus to make the following 2 classes pass with node 20 on win or Ubuntu:

- BlockBlobHighlevel
- ServiceAPIs

BTW, the failed test cases are all for "read ECONNRESET", it might caused by the keepalive timeout change might be too short.

blueww:
@Slach 

This RP looks target for a temp fix of CPP SDK issue (you said : "We are going to fix this behavior in the next version of clickhouse-server"), and this looks a high risk change since there are so many test cases failed. 
So to avoid blocking other Azurite user, we might don't prefer to take this change. 

I would suggest you contact CPP SDK team if this is a CPP SDK issue.
If you would like to use Azurite with the customized keeplivetimeout, I would suggest you build your own Azurite package from your code, instead push the change to Azurite public repo.

README.md
blueww: It looks might not be 5, since so many test cases failed with "read ECONNRESET", looks the network break since keep live timeout.

blueww:
Close this PR as its looks a temp fix for client SDK issue, and take many Azurite Unit test fail.

Slach:
@blueww 
I figure out with tests, now it passed in my local machine could we reopen PR? or i shall create new?

PR #2453, opened 2024-08-22
Prevents invalid metadata names in Blob APIs

### LINKED ISSUE #2451: PutBlob API allows invalid metadata name with hyphen
blob

latest (3.31.0)
DockerHub
Original [fix](https://github.com/Azure/Azurite/pull/1920) handled the issue within the explicit setMetadata  API, however doesn't handle the code path of setting the metadata through the [putBlob API](https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob?tabs=microsoft-entra-id#request-headers-all-blob-types) by setting x-ms-meta-name in the request header.
Call put blob with setting x-ms-meta-name to key that contains hyphen. Also see the test [here](https://github.com/Azure/Azurite/pull/1920).
If possible, please provide the debug log using the -d parameter, replacing \<pathtodebuglog\> with an appropriate path for your OS, or review the instructions for docker containers:

```
-d "<pathtodebuglog>"
```

Please be sure to remove any PII or sensitive information before sharing!  
The debug log will log raw request headers and bodies, so that we can replay these against Azurite using REST and create tests to validate resolution.

No.

benny-n:
Fixes #2451 

This PR also follows the [suggestion](https://github.com/Azure/Azurite/issues/2451#issuecomment-2301483686) to fix all the blob APIs, not just putBlob.

To make sure this is always checked, we check for valid metadata names in the conversion function instead of in each API individually.

### PR Branch Destination

### Always Add Test Cases

### Add Change Log

### Development Guideline

### DIFF
--- a/tests/blob/apis/blob.test.ts
+++ b/tests/blob/apis/blob.test.ts
@@ -496,6 +496,31 @@ describe("BlobAPIs", () => {
 
   });
 
+  it("should fail when upload has metadata names that are invalid C# identifiers @loki @sql", async () => {
+    let invalidNames = [
+      "1invalid",
+      "invalid.name",
+      "invalid-name",
+    ]
+    for (let i = 0; i < invalidNames.length; i++) {
+      const metadata = {
+        [invalidNames[i]]: "value"
+      };
+      let hasError = false;
+      try {
+        await blockBlobClient.upload(content, content.length, { metadata });
+      } catch (error) {
+        assert.deepStrictEqual(error.statusCode, 400);
+        assert.strictEqual(error.code, 'InvalidHeaderValue');
+        hasError = true;
+      }
+      if (!hasError)
+      {
+        assert.fail();
+      }
+    }
+  });
+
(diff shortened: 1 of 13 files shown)


benny-n:

tomerrt (approved):

tests/blob/apis/blob.test.ts
```diff
@@ -496,6 +496,31 @@ describe("BlobAPIs", () => {
 
   });
 
+  it("should fail when upload has metadata names that are invalid C# identifiers @loki @sql", async () => {
```
blueww, line 499: It would be great if all related API can be covered by test cases.

blueww (changes requested):
Thanks for the contribution!
The PR generally look great to me!

Only one comment for more test coverage.

benny-n:
@blueww Thanks for the review!

Added more test coverage as requested.

blueww (approved):

PR #2461, opened 2024-09-03
Download 0 size blob with range > 0 should has header "Content-Range: bytes \*/0" in returned error

blueww:
Thanks for contribution! Please go through following checklist before sending PR.

Fix https://github.com/Azure/Azurite/issues/2458

### PR Branch Destination

- For Azurite V3, please send PR to `main` branch.
- For legacy Azurite V2, please send PR to `legacy-dev` branch.

### Always Add Test Cases

Make sure test cases are added to cover the code change.

### Add Change Log

Add change log for the code change in `Upcoming Release` section in `ChangeLog.md`.

### Development Guideline

Please go to CONTRIBUTION.md for steps about setting up development environment and recommended Visual Studio Code extensions.

PR #2462, opened 2024-09-04
Ignore invalid range in blob download

### LINKED ISSUE #2458: Empty blob access error using Java SDK

blob

3.32.0

Docker

n/a

Downloading an empty (0 bytes) blob from Azurite gives a 416 error response, while Azure works as expected. With version 3.31.0, both Azurite and Azure work as expected without an error.

This seems to contradict #2410, which reported the opposite behavior for an earlier Azurite version. As a consequence, #2417 was implemented.

See https://github.com/cnaj/azurite-empty-blob-issue-demo for a reproduction of the issue.

no

cnaj:
Fix #2458, based on #2461

I'm not sure how to provoke an invalid range header in the unit tests; @blueww can you give me a hint?

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -7,7 +7,8 @@
 Blob:
 
 - Fixed an issue where all blob APIs allowed metadata names which were not valid C# identifiers.
-- Fixed issue of download 0 size blob with range > 0 should has header "Content-Range: bytes \*/0" in returned error. (issue #2458)
+- Fixed issue of download 0 size blob with range > 0 should have header "Content-Range: bytes \*/0" in returned error. (issue #2458)
+- Aligned behavior with Azure to ignore invalid range requests for blob downloads. (issue #2458)
 
 ## 2024.08 Version 3.32.0
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -7,13 +7,14 @@
 Blob:
 
 - Fixed an issue where all blob APIs allowed metadata names which were not valid C# identifiers.
+- Fixed issue of download 0 size blob with range > 0 should has header "Content-Range: bytes \*/0" in returned error. (issue #2458)
 
 ## 2024.08 Version 3.32.0
 
 General:
 
 - Bump mysql2 to resolve to 3.10.1 for security patches
-- Fixed an issue where premature client disconnections led to all following requests failing with a 500 error. (issue #1346) 
+- Fixed an issue where premature client disconnections led to all following requests failing with a 500 error. (issue #1346)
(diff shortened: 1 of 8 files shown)


cnaj:
@microsoft-github-policy-service agree company="corpuls"

blueww:
@EmmaZhu

Do you have any idea how to input a invalid range header like "0--1" in download blob API by JS SDK (in Azurite test)?

EmmaZhu:
Following is sample to set a customized header to a blobClient instance:

```
// Define a PipelinePolicy to set a customized header
export const injectorPolicyName = "injectorPolicy";

export function injectorPolicy(): PipelinePolicy {
  return {
    name: injectorPolicyName,
    async sendRequest(request: PipelineRequest, next: SendRequest): Promise<PipelineResponse> {
      request.headers.set('headername', `headervalue`);
      return next(request);
    },
  };
}
```

```
    // Insert the policy to pipeline in blobClient's contenxt
    const injector = injectorPolicy();

    const pipeline: Pipeline = (blobClient as any).storageClientContext.pipeline;
    pipeline.addPolicy(injector, {phase: 'Retry'});

   // code against blobClient
```

blueww:
@cnaj 
Would you please look at the sample code from Emma for how to add customized header when send request with JS storage SDK,  and follow up it to add the test case?

cnaj:
I'm still trying to figure out how to apply @EmmaZhu's suggestion. For instance, there is no `pipeline` property on `blobClient.storageClientContext`, as far as I can see in the debugger. Also, this is an undocumented feature for which I don't have enough information... Could you please clarify?

blueww:
@EmmaZhu 
Would you please look at the above comments, and share the info that @cnaj  needed?
Thanks!

gaul:
@EmmaZhu is there any way to unblock this PR?  I have a duplicate one open.

blueww:
@gaul , @cnaj 

This is a sample azurite test case which use "addpolicy" to add a customized policy to change to request send from storage js SDK.
Would you please refer it for how to apply the test case?

https://github.com/Azure/Azurite/blob/3e25ee56cfdf3a0b236cedb781152b1073ec41d6/tests/table/auth/tableCorsRequest.test.ts#L64

cnaj:
Thank you @blueww, this hint pointed me into the right direction. I've added an additional test as a sanity check for the new `RangePolicy`.

blueww (approved):

blueww:
Thanks @cnaj!
The PR looks good for me, and I have merged it.

Issue #2467, opened 2024-09-25
Consider removing '*' from Activation Events


JustinGrote:
The extension is very heavyweight startup, 2s on my machine and a lot of cpu cycles each time even if I'm not using it.

Ideally your activation events would be the existing onCommand events, and a `workspaceContains` even that looks for the typical azurite cache files indicating azurite is in use in the workspace.

At worst, change `*` to `onStartupFinished` so that vscode startup time is not impacted by the extension loading.

Due to this behavior I generally have to disable it and only enable it in workspaces I use it, it would be nice for it to be more behaved.

blueww:
@JustinGrote 

Thanks for the suggestion!
We will look into it.

Besides that, Azurite welcome contribution!
It would be great if you could raise a PR to fix it.

PR #2468, opened 2024-09-26
Activate only on command or Azurite file presence

JustinGrote:
Closes https://github.com/Azure/Azurite/issues/2467

## Note
Optionally this could additionally add `onStartupFinished` and that will at least minimise the performance impact of Azurite on vscode startup, however if you're going to do that, no point in having any of the other activation events and they should all be removed.

Potentially considered a breaking change because a "new" workspace won't activate the extension and people may be used to seeing the blob/table/etc. links in the bottom status tray and they won't show up until you run an azurite command. Once run however and the azurite files are present in the workspace, it will auto-start from then on for the project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,9 @@
     "vsce": "^2.7.0"
   },
   "activationEvents": [
-    "*",
+    "workspaceContains:__azurite*.json",
+    "workspaceContains:__blobstorage__",
+    "workspaceContains:__queuestroage__",
     "onCommand:azurite.start",
     "onCommand:azurite.close",
     "onCommand:azurite.clean",
--- a/package.json
+++ b/package.json
@@ -89,21 +89,7 @@
     "vsce": "^2.7.0"
   },
   "activationEvents": [
-    "workspaceContains:__azurite*.json",
-    "workspaceContains:__blobstorage__",
-    "workspaceContains:__queuestroage__",
-    "onCommand:azurite.start",
-    "onCommand:azurite.close",
-    "onCommand:azurite.clean",
-    "onCommand:azurite.start_blob",
-    "onCommand:azurite.close_blob",
-    "onCommand:azurite.clean_blob",
-    "onCommand:azurite.start_queue",
-    "onCommand:azurite.close_queue",
-    "onCommand:azurite.clean_queue",
-    "onCommand:azurite.start_table",
-    "onCommand:azurite.close_table",
-    "onCommand:azurite.clean_table"
+    "onStartupFinished"
   ],
   "contributes": {
     "commands": [


blueww:
@XiaoningLiu 
Would you please help to review this PR?

blueww:
@JustinGrote 

Thanks for the contribution!

However, this breaking is not acceptable to Azurite. 
We would expect a PR that make Azurite load faster in VSC, but not expect PR that won't load Azurite by default in VSC.

JustinGrote:
I will adjust it to OnStartupCompleted then, but this does not make you a good vscode citizen loading your extension in every workspace even when its not necessary, especially with your heavy startup times.

JustinGrote:
@blueww please re-review. Info about OnStartupFinished
https://code.visualstudio.com/api/references/activation-events#onStartupFinished

blueww:
> @blueww please re-review. Info about OnStartupFinished https://code.visualstudio.com/api/references/activation-events#onStartupFinished

Thanks @JustinGrote  for the sharing!
I see [onStartupFinished](https://code.visualstudio.com/api/references/activation-events#onStartupFinished) will be activated some time after VS Code starts up. This is similar to the * activation event, but it will not slow down VS Code startup. 

So I can understand "onStartupFinished" instead of "*" can avoid slow down VS Code startup.
However, I see many other event are removed besides than "*", will there be any side effect?

JustinGrote:
> > @blueww please re-review. Info about OnStartupFinished https://code.visualstudio.com/api/references/activation-events#onStartupFinished
> 
> Thanks @JustinGrote  for the sharing!
> I see [onStartupFinished](https://code.visualstudio.com/api/references/activation-events#onStartupFinished) will be activated some time after VS Code starts up. This is similar to the * activation event, but it will not slow down VS Code startup. 
> 
> So I can understand "onStartupFinished" instead of "*" can avoid slow down VS Code startup.
> However, I see many other event are removed besides than "*", will there be any side effect?

No side effect, the other events are redundant if you are always activating on startup. They are what you *should* be using to minimize usage, but you have chosen to always run on startup even if it is not relevant to the workspace.

blueww:
@XiaoningLiu , @EmmaZhu 

Would you please help to review?

blueww (approved):
I have checked B/T/Q in the button can be load when open VSC. 
Not see other side effect in my testing.

XiaoningLiu (approved):

blueww:
@JustinGrote 

We have approved the PR, but the test failure blocks PR merge.

Would you please rebase/merge you PR with the latest change on main, so we can include the test fix in PR validation.

JustinGrote:
I have rebased as requested. FYI in the future, you have write access to pull request branches and can perform this rebase yourself without needing to wait on me, thanks!

PR #2472, opened 2024-10-04
Only include metadata when requested

### LINKED ISSUE #1730: GetBlobsAsync always includes Metadata while real Azure doesn't
Blob
3,19,0
npm
16.15.0
When calling GetBlobsAsync (Azure .NET SDK) without specifiying the Traits, the metadata is included. This is not the case when executing against real Azure Storage, there the Metadata is empty and you must specify explicitly the BlobTraits.

1) Create a Blob in Azurite Container (named "12345") with Metadata on it
2) BlobServiceClient.GetBlobContainerClient( "12345").GetBlobsAsync(**BlobTraits.Metadata**);
Log:
`127.0.0.1 - - [26/Oct/2022:10:56:48 +0000] "GET /devstoreaccount1/12345?restype=container&comp=list&**include=Metadata** HTTP/1.1" 200 -`

Result: BlobItem has Metadata-Property filled  
Expected: This is okay and like it should be

3) BlobServiceClient.GetBlobContainerClient( "12345").GetBlobsAsync();
Log:
`127.0.0.1 - - [26/Oct/2022:11:08:56 +0000] "GET /devstoreaccount1/12345?restype=container&comp=list HTTP/1.1" 200 -`

Result: BlobItem has Metadata-Property filled  
Expected: BlobItem should NOT have Metadata-Property filled (like real Azure Storage behaves)

frederikrosenberg:
Closes #1730

Only include metadata when requested on blob list endpoints.

### DIFF
--- a/tests/blob/apis/container.test.ts
+++ b/tests/blob/apis/container.test.ts
@@ -548,12 +548,18 @@ describe("ContainerAPIs", () => {
 
   it("should correctly list all blobs in the container using listBlobFlatSegment with default parameters @loki @sql", async () => {
     const blobClients = [];
+    const metadata = {
+      keya: "a",
+      keyb: "c"
+    };
     for (let i = 0; i < 3; i++) {
       const blobClient = containerClient.getBlobClient(
         getUniqueName(`blockblob${i}/${i}`)
       );
       const blockBlobClient = blobClient.getBlockBlobClient();
-      await blockBlobClient.upload("", 0);
+      await blockBlobClient.upload("", 0, {
+        metadata
+      });
       blobClients.push(blobClient);
     }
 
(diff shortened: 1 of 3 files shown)


frederikrosenberg:

tests/blob/apis/container.test.ts
```diff
@@ -548,12 +548,18 @@ describe("ContainerAPIs", () => {
 
   it("should correctly list all blobs in the container using listBlobFlatSegment with default parameters @loki @sql", async () => {
     const blobClients = [];
+    const metadata = {
```
blueww, line 551: As 2 APIs are updated, the test should cover both 2 APIs.

blueww (changes requested):
The change generally looks good to me. 
Only a comment for testing.

blueww (approved):

Issue #2473, opened 2024-10-08
GetBlob on archive tier blob should fail


kfcss:
### Which service(blob, file, queue, table) does this issue concern?
Blob

### Which version of the Azurite was used?
3.31.0

### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
Docker. From `mcr.microsoft.com/azure-storage/azurite:3.31.0`

### What's the Node.js version?
As in the container. It is v20.10.0.

### What problem was encountered?
GetBlob on an archive tier blob succeeded. It should fail.

### Steps to reproduce the issue?

1. Upload a block blob
2. Change tier to archive
3. GetBlob succeeds, which it should not according to the documentation.

> A Get Blob operation on an archived block blob will fail.
Ref: https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob#remarks

The debug log contains request made with Azure Storage Explorer and might contain requests for listing blobs as well.
[debug.log](https://github.com/user-attachments/files/17288544/debug.log)

### Have you found a mitigation/solution?
No

blueww:
@kfcss 

Thanks for raising this issue!
It looks archive tier fail request still no implemented in Azurite.

Azurite welcome contribution!
It would be great if you could raise a PR to fix this.

jainakanksha-msft:
This fix has been merged, so I’m closing this issue.

PR #2483, opened 2024-10-24
Added accessTier check to download call for BlobHandler.ts to fail on…

BrendonK:
… archived blobs.

#2473

### PR Branch Destination

### Always Add Test Cases

### Add Change Log

### Development Guideline

### DIFF
--- a/src/blob/handlers/BlobHandler.ts
+++ b/src/blob/handlers/BlobHandler.ts
@@ -80,6 +80,10 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
       options.modifiedAccessConditions
     );
 
+    if (blob.properties.accessTier === Models.AccessTier.Archive) {
+      throw StorageErrorFactory.getBlobArchived(context.contextId!);
+    }
+
     if (blob.properties.blobType === Models.BlobType.BlockBlob) {
       return this.downloadBlockBlobOrAppendBlob(options, context, blob);
     } else if (blob.properties.blobType === Models.BlobType.PageBlob) {
--- a/tests/blob/apis/blob.test.ts
+++ b/tests/blob/apis/blob.test.ts
@@ -253,7 +253,7 @@ describe("BlobAPIs", () => {
     assert.fail();
   });
 
-  it("download should not work when blob in Archive tier", async () => {
+  it("download should not work when blob in Archive tier @loki @sql", async () => {
     try {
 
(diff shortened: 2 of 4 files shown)


BrendonK:

src/blob/handlers/BlobHandler.ts
```diff
@@ -80,6 +80,10 @@ export default class BlobHandler extends BaseHandler implements IBlobHandler {
       options.modifiedAccessConditions
     );
 
+    if (blob.properties.accessTier === Models.AccessTier.Archive) {
+      throw StorageErrorFactory.getBlobArchived(context.contextId!);
```
blueww, line 84: Please make sure the error returned here is same as the error return from product Azure server. (status code, error code, error message.)

BrendonK: @blueww I got the error codes/message from [here](https://learn.microsoft.com/en-us/rest/api/storageservices/blob-service-error-codes) so it should be okay I think:
BlobArchived	Conflict (409)	This operation is not permitted on an archived blob.

BrendonK: @blueww some tests are failing - I will check the failures to see what's causing it.

blueww (approved):

tests/blob/apis/blob.test.ts
```diff
@@ -253,6 +253,22 @@ describe("BlobAPIs", () => {
     assert.fail();
   });
 
+  it("download should not work when blob in Archive tier", async () => {
```
blueww, line 256: Please add `@loki @sql` as other test cases

tests/blob/sas.test.ts
```diff
@@ -2122,7 +2122,7 @@ describe("Shared Access Signature (SAS) authentication", () => {
     }
     assert.ok(error !== undefined);
     assert.equal(error.statusCode, 409);
-    assert.equal(error.details.code, "BlobArchived");
+    assert.equal(error.details.code, "CannotVerifyCopySource");
```
blueww, line 2125: The SQL test case failure on this, looks related with `sourceBlob.setAccessTier("Archive");` on line 2113 don't have await, it's async call, without await the access tier might still not convert to archive.

Add await should can resolve the failure.
 `await sourceBlob.setAccessTier("Archive");`

blueww: I have fixed the case

PR #2486, opened 2024-11-09
Handle getting zero-length blobs with `Content-Range`

gaul:
This behavior matches Azure.  AWS Java SDK probes for a zero-length blob when a request with a Content-Range fails.

### DIFF
--- a/src/blob/errors/StorageError.ts
+++ b/src/blob/errors/StorageError.ts
@@ -29,7 +29,8 @@ export default class StorageError extends MiddlewareError {
     storageErrorCode: string,
     storageErrorMessage: string,
     storageRequestID: string,
-    storageAdditionalErrorMessages: { [key: string]: string } = {}
+    storageAdditionalErrorMessages: { [key: string]: string } = {},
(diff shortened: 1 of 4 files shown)


gaul:

gaul:
This is difficult to add a test for since the AWS JavaScript SDK doesn't allow sending a zero `count`:

```javascript
          range: offset === 0 && !count ? undefined : rangeToString({ offset, count }),
```

https://github.com/Azure/azure-sdk-for-js/blob/fe02d22be5944988b89f22421c0dcc428c1ae525/sdk/storage/storage-blob/src/Clients.ts#L1235

gaul:
See related Azure/azure-sdk-for-java#42870.

blueww:
The change of this PR is almost covered by https://github.com/Azure/Azurite/pull/2462.
Besides that, Azurite unit test cases based on JS SDK, which support send 0 range. See the test case change in https://github.com/Azure/Azurite/pull/2462

blueww:
Close as the PR https://github.com/Azure/Azurite/pull/2462 merged, which covers fix of this PR.

PR #2487, opened 2024-11-10
Check both Content-MD5 and x-ms-blob-content-md5

gaul:


blueww (changes requested):
Please add a test case to cover it.
Besides that, please consider that API which support Content-MD5 and x-ms-blob-content-md5.

gaul:
Could you expand on your suggestion?  I believe that for block blobs `Content-MD5` and `x-ms-blob-content-md5` are synonyms: https://technet2.github.io/Wiki/blogs/windowsazurestorage/windows-azure-blob-md5-overview.html .  It appears that the JavaScript API only sends the former but the Java API sends the latter.  How can I inject a header that is not have a `x-ms-meta-` prefix?

gaul:
@blueww Is there any way to move this PR forward?  I don't believe that the test you suggested is possible with the SDK.

blueww:
@gual

Please refer this test case for how to add header to request send from Storage JS SDK with customized policy.

https://github.com/Azure/Azurite/blob/3e25ee56cfdf3a0b236cedb781152b1073ec41d6/tests/table/auth/tableCorsRequest.test.ts#L64

The customized policy class is defined in https://github.com/Azure/Azurite/blob/main/tests/blob/RequestPolicy/OPTIONSRequestPolicyFactory.ts

The sample test case changes several items in the request, if only for header, see a more simple guild in https://github.com/Azure/Azurite/pull/2462#issuecomment-2340150113

gaul:
@blueww I added the suggested test -- please take a look.

tests/blob/RequestPolicy/CustomMD5PolicyFactory.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  constructor(nextPolicy: any, options: any, md5: any) {
+    super(nextPolicy, options);
+    this.md5 = md5;
+  }
+
+  // Customize HTTP requests and responses by overriding sendRequest
+  // Parameter request is Azure.WebResource type
+  async sendRequest(request: WebResource) {
+    request.headers.set("x-ms-blob-content-md5", `${this.md5}`);
```
blueww, line 30: It would be better to write a more generic policy to add a header from input header name and value, then other coming test cases can also use it, instead of each header has seperate class to add header.

blueww (approved):

PR #2526, opened 2025-01-22
Remove unneeded `multistream` dependency

talentlessguy:
Removed dependency on multistream which is no longer needed.

`stream.pipeline` has been available since Node 10, which is the lowest version Azurite supports.


talentlessguy:

Issue #2525, opened 2025-01-14
--oauth with no 'basic' causes crash


voltagex:
```
Now using node v23.6.0 (npm v10.9.2)
$ npx azurite
(node:32623) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Azurite Blob service is starting at http://127.0.0.1:10000
Azurite Blob service is successfully listening at http://127.0.0.1:10000
Azurite Queue service is starting at http://127.0.0.1:10001
Azurite Queue service is successfully listening at http://127.0.0.1:10001
Azurite Table service is starting at http://127.0.0.1:10002
Azurite Table service is successfully listening at http://127.0.0.1:10002

$ npx azurite --oauth
(node:32664) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Exit due to unhandled error: this.oauth.toLowerCase is not a function

$ npx --version
10.9.2

```

blueww:
@voltagex 

Thanks for reporting this issue!

Please use "--oauth basic", the parameter "--oauth" must has a value.
For the crash, as this is a invalid usage, this might not be our priority recently.

PR #2554, opened 2025-05-21
Throwing error message if oauth argument set to anything other than basic for usse #2525

matt-lethargic:
This pull request introduces stricter validation for the `oauth` flag across multiple environment classes and improves type safety in the `ConfigurationBase` class as highlighted in #2525 . The main changes ensure that only the value `"basic"` is accepted for the `oauth` flag, with an appropriate error thrown for invalid values.

### Stricter validation for `oauth` flag:

* [`src/blob/BlobEnvironment.ts`](diffhunk://#diff-56fdff4a5dc4d9d8f36522bb4a24fd302eebae9f19716ff4f5ca8d05ba8bd4feR135-R143): Added validation in the `oauth` method to ensure that only `"basic"` is accepted as a valid value for the `oauth` flag, throwing a `RangeError` otherwise.
* [`src/common/Environment.ts`](diffhunk://#diff-228379cf1c30321ce7c0346d6cf069a6df7b65896326ef717d1b152431d21b35R204-R212): Applied the same validation logic in the `oauth` method for the `Environment` class.
* [`src/queue/QueueEnvironment.ts`](diffhunk://#diff-9eea57d39c1deb9efc262355f2319dbd530490fc2ca36dca032fc228d1d9dd94R119-R127): Implemented the same validation in the `oauth` method for the `QueueEnvironment` class.
* [`src/table/TableEnvironment.ts`](diffhunk://#diff-455d420138770ed90dfd6fb548b76013bd2bc67992f0ec1480a4f9c6c95b13a0R156-R164): Added the same validation in the `oauth` method for the `TableEnvironment` class.

### Improved type safety:

* [`src/common/ConfigurationBase.ts`](diffhunk://#diff-ad5434dcf8c1ef2e3e062017e76ba23ed5830b56aa24f66dd50130287b0911c6L94-R94): Updated the `getOAuthLevel` method to explicitly check that `this.oauth` is a string before processing it, improving type safety.

### DIFF
--- a/src/blob/BlobEnvironment.ts
+++ b/src/blob/BlobEnvironment.ts
@@ -132,7 +132,14 @@ export default class BlobEnvironment implements IBlobEnvironment {
   }
 
   public oauth(): string | undefined {
-    return this.flags.oauth;
+    if (this.flags.oauth !== undefined) {
+      if (this.flags.oauth !== "basic") {
+        throw RangeError(
+          `Must provide a valid value for parameter --oauth, only 'basic' is supported.`
+        );
+      }
+      return this.flags.oauth;
+    }
   }
 
   public disableProductStyleUrl(): boolean {
--- a/src/common/ConfigurationBase.ts
+++ b/src/common/ConfigurationBase.ts
@@ -91,7 +91,7 @@ export default abstract class ConfigurationBase {
   }
 
   public getOAuthLevel(): undefined | OAuthLevel {
-    if (this.oauth) {
+    if (this.oauth && typeof this.oauth === "string") {
       if (this.oauth.toLowerCase() === "basic") {
         return OAuthLevel.BASIC;
       }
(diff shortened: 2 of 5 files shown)


src/blob/BlobEnvironment.ts
```diff
@@ -132,7 +132,14 @@ export default class BlobEnvironment implements IBlobEnvironment {
   }
 
   public oauth(): string | undefined {
-    return this.flags.oauth;
+    if (this.flags.oauth !== undefined) {
+      if (this.flags.oauth !== "basic") {
```
blueww, line 136: Please don't hard code "basic".
The basic here means Azurite only do basic validation of the Bearer token.   It give the scalability for adding more validation to the bearer token in the future. 
So please add an enum for the value support by Oauth, and report error when the value not in the enum.

src/blob/BlobEnvironment.ts
```diff
@@ -132,7 +132,14 @@ export default class BlobEnvironment implements IBlobEnvironment {
   }
 
   public oauth(): string | undefined {
-    return this.flags.oauth;
+    if (this.flags.oauth !== undefined) {
```
blueww, line 135: Please add changelog, and better with test cases.

PR #2581, opened 2025-08-07
Change character validation in tag key to align with server behavior

EmmaZhu:

### PR Branch Destination

### Always Add Test Cases

### Add Change Log

### Development Guideline

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,10 @@
 
 ## Upcoming Release
 
+Blob:
+
+- Changed charaters validation against tag key in filtering blobs with tags to align with service side behavior. (issue #2561)
+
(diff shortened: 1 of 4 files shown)


SamPersson:
It seems the failing test is correct: `key111==value1` should still result in http status 400

> If tag names are regular SQL identifiers, they may be present without escaping; if they contain any special characters, they must be delimited with double quotes (e.g. "TagName" = 'TagValue').

(https://learn.microsoft.com/en-us/rest/api/storageservices/specifying-conditional-headers-for-blob-service-operations#tags-predicate-syntax)

So I guess it needs separate validation for double quote delimited tag keys, and only then allow `+-./:=`

Issue #2589, opened 2025-08-28
Azurite emulator does not return 412 Precondition failed providing IfMatch = "*" for blob uploads for non-existent blob


jasoncheungg:
**Which service(blob, file, queue, table) does this issue concern?**
blob

**Which version of the Azurite was used?**
3.34.0

**Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)**
Visual Studio [here](https://learn.microsoft.com/en-us/azure/storage/common/storage-install-azurite?toc=%2Fazure%2Fstorage%2Fblobs%2Ftoc.json&bc=%2Fazure%2Fstorage%2Fblobs%2Fbreadcrumb%2Ftoc.json&tabs=visual-studio%2Cblob-storage#:~:text=C%3A%5CProgram%20Files%5CMicrosoft%20Visual%20Studio%5C2022%5CEnterprise%5CCommon7%5CIDE%5CExtensions%5CMicrosoft%5CAzure%20Storage%20Emulator)

**What's the Node.js version?**
v22.17.0

**What problem was encountered?**
Expected blob upload operation to fail with 412 Precondition failed when the blob does not already exist and the If-Match header is set to "*". However, the operation succeeds instead of a 412 precondition failure raising which is not consistent with Azure Storage and the Azure Storage Emulator (Legacy) here.

**Steps to reproduce the issue?**
1. Start the Azurite emulator
2. Upload a blob with the if-match="*" header
3. Observe that the request succeeds which is incorrect instead of returning a 412 precondition failed error

**Have you found a mitigation/solution?**
No

jainakanksha-msft:
@jasoncheungg , This fix has been merged, so I’m closing this issue.

PR #2621, opened 2025-12-23
Fixes Azurite V3 emulator not returning 412 Precondition Failed if a resource (blob) does not exist and "*" IfMatch is provided.

jasoncheungg:
This PR fixes the issue where the Azurite V3 emulator does not return 412 Precondition Failed if a resource (blob) does not exist and "*" IfMatch is provided.

Issue #2589

### DIFF
--- a/src/blob/conditions/WriteConditionalHeadersValidator.ts
+++ b/src/blob/conditions/WriteConditionalHeadersValidator.ts
@@ -102,10 +102,8 @@ export default class WriteConditionalHeadersValidator
       if (conditionalHeaders.ifMatch && conditionalHeaders.ifMatch.length > 0) {
         // If a request specifies both the If-Match and If-Unmodified-Since headers,
         // the request is evaluated based on the criteria specified in If-Match.
-        if (conditionalHeaders.ifMatch[0] !== "*") {
-          throw StorageErrorFactory.getConditionNotMet(context.contextId!);
-        }
-        return;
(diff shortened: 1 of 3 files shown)


jasoncheungg:
@microsoft-github-policy-service agree company="Microsoft"

src/blob/conditions/WriteConditionalHeadersValidator.ts
```diff
@@ -102,10 +102,8 @@ export default class WriteConditionalHeadersValidator
       if (conditionalHeaders.ifMatch && conditionalHeaders.ifMatch.length > 0) {
         // If a request specifies both the If-Match and If-Unmodified-Since headers,
         // the request is evaluated based on the criteria specified in If-Match.
-        if (conditionalHeaders.ifMatch[0] !== "*") {
-          throw StorageErrorFactory.getConditionNotMet(context.contextId!);
-        }
-        return;
+        // Throw if there is any value in if-match for non exist blob
```
blueww, line 105: Could you validate all the write request call this function on real azure server will have this behavior? 
If so, this change is OK.
Else, you might need to so some check for the API here.

Others are all good for me.

jasoncheungg: @blueww If the If-Match = "*" and the resource does not exist, then it will also throw 412s.

blueww (approved):

PR #2624, opened 2026-01-16
Modernize and streamline status bar visual appearance

JustinGrote:
This Pull Request optimizes the status bar to use icons rather than text, moving that status to the tooltip, and provides animations when starting and stopping.

This provides a much cleaner experience for the average user and no longer takes up half of the status bar to do so.

https://github.com/user-attachments/assets/93cecdd2-212b-4842-bda7-432fb0d6f80a

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,6 +8,7 @@ General:
 
 - Performance improvements for internal metadata access using in-memory metadata store
 - Fix building failure on Node 22 platform.
+- Modernize and improve status bar icons
 
 ## 2025.07 Version 3.35.0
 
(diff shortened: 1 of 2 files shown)


JustinGrote:
I also think that not everything logged to info should have a popup, it is very distracting, but that will be in another PR.

blueww:
@XiaoningLiu , @EmmaZhu  How do you think about this change?

@JustinGrote , I am l little worry about customer can't know what's the icon means, and need to check each icon to find the one he/she want.

JustinGrote:
@blueww I'm happy to add the words "Blob", "Queue", "Table" next to the icons, however if you hover over any of the icons it says exactly what it is for. I can add "Azurite Blob Storage Emulator" or something more specific for the hover if you would prefer.

Please note other icons like github, alert, etc. all do it with just the icon and users seem to be just fine with it :)

jainakanksha-msft:
@JustinGrote  , could you please refresh your PR with main, and address the review comments if any to move this PR forward.

PR #2636, opened 2026-03-17
Fix FilterBlobs crash when where clause is omitted

qbradley:
Service_FilterBlobs and Container_FilterBlobs crashed with HTTP 500 when the optional 'where' query parameter was not provided. The response object used 'options.where!' (non-null assertion) which set the required 'where' field to undefined, causing serialization to fail.

Replace 'options.where!' with 'options.where || ""' so that a missing where clause produces an empty string instead of undefined. The handler now returns 200 with an empty blob list when no filter expression is given.

Add regression test that sends a raw HTTP FilterBlobs request without a where parameter and asserts both service-level and container-level endpoints return 200.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -9,6 +9,7 @@ General:
 - Performance improvements for internal metadata access using in-memory metadata store
 - Fix building failure on Node 22 platform.
 - Fix * IfMatch for non-existent resource not throwing 412 Precondition Failed
+- Fix FilterBlobs returning 500 instead of 200 with optional missing where parameter.
 
 ## 2025.07 Version 3.35.0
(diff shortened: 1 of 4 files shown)


jainakanksha-msft:
@qbradley, could you please refresh your PR with main, and address the review comments to move this PR forward.

jainakanksha-msft:
Superseded by #2782, rebased on current `main`. The replacement keeps the handler fix and addresses the review feedback by using the SDK authenticated protocol operations (no hand-rolled SharedKey signing or hardcoded API version), asserting both an empty `where` value and empty blob results, and clarifying the changelog entry.

PR #2638, opened 2026-04-02
SFI - table tests

jwdb:
I've updated the azure data table tests on the SFI branch from EmmaZhu with the new package. Hopefully this can get the PR merged and the api version can soon be updated again.


jainakanksha-msft:
Changes are already incorporated in https://github.com/Azure/Azurite/pull/2647

Issue #2347, opened 2024-01-23
Cannot connect to emulator in azure storage explorer when only running table service


smargoli2:
### Which service(blob, file, queue, table) does this issue concern?
Table
### Which version of the Azurite was used?
Latest
### Where do you get Azurite? (npm, DockerHub, NuGet, Visual Studio Code Extension)
npm, running with docker
### What problem was encountered?
When I run `docker run -p 10002:10002 mcr.microsoft.com/azure-storage/azurite azurite-table --tableHost 0.0.0.0 --tablePort 10002` in the vs code powershell cli, the output I see looks like this:
```
Azurite Table service is starting on 0.0.0.0:10002
Azurite Table service successfully started on 0.0.0.0:10002
```
Notice it does not say 'Azurite Table service listening on 0.0.0.0:10002'. When I try to view the emulator in Azure Storage Explorer, it says 'You are attempting to access an emulator storage account without a running emulator. Storage Explorer does not ship with an emulator, you must install your own. It is recommended that you use Azurite. Once you have installed an emulator you can start it and refresh this node.'
![image](https://github.com/Azure/Azurite/assets/22183289/a228d488-abc0-4ac7-83c2-e6daf7fb5fb5)

### Steps to reproduce the issue?
Do what I did.
If possible, please provide the debug log using the -d parameter, replacing \<pathtodebuglog\> with an appropriate path for your OS, or review the instructions for docker containers:

```
-d "<pathtodebuglog>"
```

Please be sure to remove any PII or sensitive information before sharing!  
The debug log will log raw request headers and bodies, so that we can replay these against Azurite using REST and create tests to validate resolution.

### Have you found a mitigation/solution?
Nope. I just run all the services together, but I really only need the table service.

EmmaZhu:
Hi @smargoli2 , 

I can connect to Azurite with Storage Explorer, with following steps:
1. start Azurite with only table service with command like:
  `docker run -p 10002:10002 mcr.microsoft.com/azure-storage/azurite azurite-table --tableHost 0.0.0.0 --tablePort 10002`
2.In storage explorer, add connection to Azurite's table service:
![image](https://github.com/Azure/Azurite/assets/6396650/f6b03709-2f19-42ab-a0c1-25abf87155b9)

3. Clear blobs port and queue port to avoid to connect to blob or queue service which is not started in Azurite:
[...]

Issue #145, opened 2019-01-16
Allow server instances to run on OS-assigned unused ports


poveden:
Node.js' [`Server.listen()`](https://nodejs.org/dist/latest-v10.x/docs/api/net.html#net_server_listen_port_host_backlog_callback) allows passing `0` as a port number, so the operating system assigns an arbitrary unused port. However, the current `env.init()` implementation won't allow it as `0` is "falsy":

https://github.com/Azure/Azurite/blob/f3dccf71dae73d1af8bc355e39c215515a322d4f/lib/core/env.js#L40-L42

Allowing a `0` to be passed and used would allow scenarios where fixed port number availability is not known, or where multiple, parallelized instances are being run (e.g. test suites).

A possible solution would be creating a helper function:

```js
function normalizePort(port, defaultPort) {
  if (port === 0) {
    return port;
  }

  return defaultPort;
}
```

and change the above lines to:

```js
this.blobStoragePort = normalizePort(options.p || options.blobPort, 10000);
this.queueStoragePort = normalizePort(options.q || options.queuePort, 10001);
this.tableStoragePort = normalizePort(options.t || options.tablePort, 10002);
```

XiaoningLiu:
That sounds good, can you make a PR?

kfarmer-msft:
Out of curiosity, how would a client discover the port number for use in a test?

XiaoningLiu:
@kfarmer-msft  Default ports will be used without specified configurations when starting Azurite.

poveden:
> Out of curiosity, how would a client discover the port number for use in a test?

@kfarmer-msft While running from the command line, the output will tell you the assigned ports:

```shell
$ azurite -p 0 -q 0 -t 0

 _______                   _
(_______)                 (_)  _
 _______ _____ _   _  ____ _ _| |_ _____
|  ___  (___  ) | | |/ ___) (_   _) ___ |
| |   | |/ __/| |_| | |   | | | |_| ____|
|_|   |_(_____)____/|_|   |_| \__)_____)

Azurite, Version 2.7.0
A lightweight server clone of Azure Storage

Azure Table Storage Emulator listening on port 55292
Azure Queue Storage Emulator listening on port 55293
Azure Blob Storage Emulator listening on port 55294
```

If started programmatically, calling Node.js' [`Server.address()`](https://nodejs.org/docs/latest/api/net.html#net_server_address) will give you the bound endpoint:

```js
const queue = new AzuriteQueue();
[...]

PR #2643, opened 2026-04-27
Fix azurite-table banner reporting configured port instead of bound port

### LINKED ISSUE #2642: azurite-table reports configured port (0) in startup banner instead of OS-bound port
### Which service(s)
Table

### Which version of Azurite was used
Latest `main` (verified at commit b6298b2)

### Where do you get Azurite
npm (`azurite-table` standalone entry point)

### What problem was encountered

When the standalone `azurite-table` entry point is run with `--tablePort 0` (asking the OS to assign an unused port — supported since #145), the startup banner reports `0` instead of the actual bound port:

```
Azurite Table service is starting on 127.0.0.1:0
Azurite Table service successfully started on 127.0.0.1:0    ← reports configured port, not bound port
```

The blob and queue entry points correctly use `server.getHttpServerAddress()` for the "after start" message and report the real bound address, e.g.:

```
Azurite Blob service successfully listens on http://127.0.0.1:57443
Azurite Queue service successfully listens on http://127.0.0.1:57444
```

This breaks any consumer that parses the banner to discover where the table service is listening — including:

- The dynamic-port scenario originally requested in #145 (test harnesses spinning up parallel Azurite instances).
- Tooling like Storage Explorer, which appears to be related to #2347 ("Cannot connect to emulator in azure storage explorer when only running table service") — the table-only banner is the only one whose wording differs from blob/queue.

### Root cause

`src/table/main.ts` builds the banner from configuration values, before `server.start()` runs:

https://github.com/Azure/Azurite/blob/b6298b2/src/table/main.ts#L62-L63

```ts
const beforeStartMessage = `Azurite Table service is starting on ${config.host}:${config.port}`;
const afterStartMessage = `Azurite Table service successfully started on ${config.host}:${config.port}`;
```

`src/blob/main.ts` and `src/queue/main.ts` instead inline the message at print time and use `server.getHttpServerAddress()`, which returns the real bound address after `listen()` resolves:

https://github.com/Azure/Azurite/blob/b6298b2/src/blob/main.ts#L41-L47
https://github.com/Azure/Azurite/blob/b6298b2/src/queue/main.ts#L87-L93

The `TableServer.afterStart()` hook (`src/table/TableServer.ts:104-107`) already logs the correct value via `logger.info`, but `main.ts` prints the wrong one to stdout.

### Steps to reproduce

```bash
node dist/src/table/main.js --inMemoryPersistence --silent --tablePort 0
```

Expected:
```
Azurite Table service is starting on 127.0.0.1:0
Azurite Table service successfully listens on http://127.0.0.1:<actual-port>
```

Actual:
```
Azurite Table service is starting on 127.0.0.1:0
Azurite Table service successfully started on 127.0.0.1:0
```

### Have you found a mitigation/solution

Yes — fix submitted as a PR which mirrors the inline `getHttpServerAddress()` pattern from `blob/main.ts` and `queue/main.ts`.

### Related issues

- #145 — original request to support OS-assigned (port 0) listening; the bug here defeats that scenario for table-only.
- #2347 — Storage Explorer cannot connect when running table-only; the inconsistent banner wording (`started on` vs blob/queue's `listens on`) is the most visible symptom of the same code path divergence and may be related.

robertmclaws:
## Summary

- The standalone `azurite-table` entry point printed its "successfully started" banner using `config.port` evaluated before `server.start()`, so callers passing `--tablePort 0` saw `port: 0` instead of the actual OS-assigned port.
- `azurite-blob` and `azurite-queue` already inline `server.getHttpServerAddress()` after `await server.start()`; this PR brings table to parity.
- Wording is also normalized from "successfully started on" to "successfully listens on" to match the sibling providers.

Fixes #2642. Related: #145 (the original OS-assigned port feature this bug defeated for table-only callers), #2347 (table-only Storage Explorer connectivity — the divergent banner wording is the most visible symptom of the same code-path inconsistency).

## Diff

```ts
// src/table/main.ts (before)
const beforeStartMessage = `Azurite Table service is starting on ${config.host}:${config.port}`;
const afterStartMessage = `Azurite Table service successfully started on ${config.host}:${config.port}`;
...
console.log(beforeStartMessage);
await server.start();
console.log(afterStartMessage);  // evaluated before start — port still 0
```

```ts
// src/table/main.ts (after — mirrors blob/main.ts and queue/main.ts)
console.log(`Azurite Table service is starting on ${config.host}:${config.port}`);
await server.start();
console.log(`Azurite Table service successfully listens on ${server.getHttpServerAddress()}`);
```

## Test plan

- [x] `npm run lint` passes
- [x] `npm run build` (tsc) passes
- [x] Manual smoke test with `--tablePort 0`:
  ```
  Azurite Table service is starting on 127.0.0.1:0
  Azurite Table service successfully listens on http://127.0.0.1:57443
  ```
- [x] Manual smoke test with `--tablePort 12345` (explicit port still works):
  ```
  Azurite Table service is starting on 127.0.0.1:12345
  Azurite Table service successfully listens on http://127.0.0.1:12345
  ```
- [x] No tests reference the old `"successfully started on"` wording (`grep -rn "successfully started" tests/` is empty for this string).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -10,6 +10,10 @@ General:
 - Fix building failure on Node 22 platform.
 - Fix * IfMatch for non-existent resource not throwing 412 Precondition Failed
 
+Table:
+
+- Fix `azurite-table` startup banner reporting the configured port (e.g. `0` when using OS-assigned ports) instead of the actual bound address. Now uses `server.getHttpServerAddress()` to match `azurite-blob` and `azurite-queue`.
+
 ## 2025.07 Version 3.35.0
 
 General:
--- a/src/table/main.ts
+++ b/src/table/main.ts
@@ -59,15 +59,13 @@ async function main() {
   // Create server instance
   const server = new TableServer(config);
 
-  const beforeStartMessage = `Azurite Table service is starting on ${config.host}:${config.port}`;
-  const afterStartMessage = `Azurite Table service successfully started on ${config.host}:${config.port}`;
   const beforeCloseMessage = `Azurite Table service is closing...`;
   const afterCloseMessage = `Azurite Table service successfully closed`;
 
   // Start Server
-  console.log(beforeStartMessage);
+  console.log(`Azurite Table service is starting on ${config.host}:${config.port}`);
   await server.start();
-  console.log(afterStartMessage);
+  console.log(`Azurite Table service successfully listens on ${server.getHttpServerAddress()}`);
   
   AzuriteTelemetryClient.init(location, !env.disableTelemetry(), env);
   await AzuriteTelemetryClient.TraceStartEvent("Table");


jainakanksha-msft:
@robertmclaws , could you please refresh your PR with main, and address the review comments if any to move this PR forward.

jainakanksha-msft (approved):

PR #2725, opened 2026-08-12
Fix serialize-javascript DoS (GHSA-qj8w-gfj5-8c6v) and simplify lockfile regression tests

jainakanksha-msft:
## What
- Relax the `serialize-javascript` override from the exact `7.0.3` pin to `^7.0.7` (resolves to 7.1.0), remediating GHSA-qj8w-gfj5-8c6v (CPU-exhaustion DoS, affects 5.0.0 - 7.0.4). The `mocha`-pulled transitive copy was stuck below the fix because of the exact pin.
- `npm audit fix`: bump lockfile-resolved `mocha` `12.0.0-rc.5` → `12.0.0-rc.6`, dedupe transitive `serialize-javascript`/`iconv-lite`.
- `tests/packageScripts.test.ts`: fix `overrides` typed as required while used optionally throughout the file.\

## Why
`serialize-javascript@7.0.3` was an exact pin, so `npm audit`/`npm audit fix` could never move the tree onto the fix even automatically — the vulnerability would have silently persisted across releases. Scope is limited to the dev/test toolchain (mocha's dependency graph); not shipped in the published `azurite` runtime.


jainakanksha-msft:
@copilot
1. if there is any build issue fix that
2. Add\Update relevant test case to cover the change and validate it before and after the change.
3. Run through test cases and validate the changes fully if it impacts the code
5. Update the Changelog file
6. Review and validate any comments(normal\suppressed) if any.
7. Rate your overall work
8. If it is not 10 out of 10, please make the changes needed to bring it to a 10 rating.
9. Share why this whole change is needed, what is the impact in details. what could happen if we didn't have this?

PR #2734, opened 2026-08-13
feat(blob): blob versioning support, opt-in per storage account (#665)

### LINKED ISSUE #665: Enable Blob versioning

gavin-thompson-postman:
## Summary

Implements **blob versioning** for the LokiJS metadata store, opt-in per storage account and **off by default**. Fixes #665.

- Versioning is an **ARM management plane** setting in Azure Storage (`Microsoft.Storage/storageAccounts/blobServices/default#isVersioningEnabled`), not part of the data plane REST API Azurite emulates — `Set Blob Service Properties` has no `Versioning` element. Since Azurite has no management plane, the setting is supplied at start up via two mutually exclusive options, `--accountConfigFile <path>` and `--accountConfig <json>`, mirrored by VS Code settings `azurite.accountConfigFile` / `azurite.accountConfig`. Multiple accounts are supported; an account that is not listed keeps the defaults, so this only ever opts accounts in.
- **No generated artifacts are edited.** The existing swagger already defines the `versionid` query parameter on 15 operations and `VersionId` / `IsCurrentVersion` / `HasVersionsOnly` on `BlobItemInternal`, so the wire contract needed no codegen changes.
- Version-creating operations follow the observed service behaviour: Put Blob, Put Block List, Copy Blob, Set Blob Metadata, Page/Append Blob Create, and Snapshot Blob create versions and return `x-ms-version-id`. Put Page and Append Block correctly do **not**.
- Reads and deletes accept `?versionid=`; `x-ms-is-current-version` is returned only for the current version. `List Blobs` supports `include=versions`. Tags and access tier are addressable per version. A version is restored by copying it over the current version.
- `Delete Blob` without a version ID turns the current version into a previous version and retains it. Deleting the **current** version by ID is refused with 403 `OperationNotAllowedOnRootBlob`, matching the service.
- Versioning can be turned on and off against an existing workspace, as on a real account. Turning it off keeps existing versions listed, readable and deletable by version ID.
- Account configuration lives in a standalone `src/common/account/` module with its own database file, so the queue and table services can share it rather than reading data owned by the blob service.

## Implementation notes

- Previous versions share the base blob's `snapshot` value in the blobs collection, so every lookup meaning "the blob itself" excludes them. Blobs written before versioning was enabled have no `isCurrentVersion` field, so those queries match `isCurrentVersion !== false` and keep resolving unchanged. `versionId` is part of the collection index, and `ensureIndex` runs for workspaces created before versioning existed.
- Retention is a property of the document, not of the account setting: an existing current blob is demoted to a previous version whenever it **is itself a version**, and removed outright only when it is not. That one rule covers Put Blob, Put Block List, Copy Blob and Delete Blob, and is why turning versioning off does not destroy history.
- **List continuation tokens** now carry a secondary key, because every version of a blob shares its name — resuming from a name alone skipped the remaining versions of the blob a page stopped inside. Tokens without a secondary key keep the historical plain-blob-name format, so listings that do not involve versions are unchanged and tokens issued by earlier versions of Azurite remain valid.
- Version IDs match the service format (RFC 3339, 7-digit fractional seconds) and are generated with a single query for the newest existing version rather than probing candidate values.
- **Loose mode** ignores a malformed or mutually exclusive `versionid` and serves the current version, rather than rejecting it — `versionid` was ignored entirely before versioning existed, so loose mode does not become stricter.
- Versioning is **not** implemented for the SQL metadata store: configuring it with `AZURITE_DB` fails at start up, and a `versionid` request returns a not-implemented error rather than silently reading the current version.

## Test plan

- [x] Full LokiJS suite: **920 passing, 4 pending** (`--grep @loki`)
- [x] `npm run test:upgrade` (upgrade/persistence compatibility): **10 passing**
- [x] `--inMemoryPersistence` mode: versioning suite passes, no database file written
- [x] `npx eslint src/**/*.ts`: clean
- [x] New suites: `blob.versioning.test.ts`, `blob.versioning.hierarchy.test.ts` (delimiter × versions, including page size 1 and a blob whose name is also another blob's prefix), `blob.versioning.toggle.test.ts`, `tests/common/AccountModel.test.ts`, `tests/common/LokiAccountModelStore.test.ts`
- [x] README updated with the new CLI flags and VS Code settings, `Blob Versions` moved out of the unsupported list, and a design doc added at `docs/designs/blob-versioning.md`
- [x] ChangeLog updated in `Upcoming Release`

### Verified against a real storage account

The versioning suites run unchanged against real Azure via the existing `AZURITE_LIVE_TEST_CONNECTION_STRING` live mode (**33 passing** against a GPv2 account with versioning enabled; the versioning-disabled and loose-mode suites are skipped there). Setup is documented in the design doc.

This found four places where the implementation had followed the prose documentation but the service behaves differently, all now following the observed behaviour:

| | Documentation implies | Service does |
| --- | --- | --- |
| Set Blob Properties | creates a version (every block blob write except Put Block) | creates **no** version, returns no `x-ms-version-id` |
| Delete current version by ID | permitted | 403 `OperationNotAllowedOnRootBlob` |
| `?snapshot=` + `?versionid=` | — | 400 `MutuallyExclusiveQueryParameters` |
| `HasVersionsOnly` | — | not reported under `include=versions`, only `include=deletedwithversions` |

## Notes

- Existing behaviour is unchanged unless an account is explicitly configured with versioning enabled.
- **One open design question for the maintainers.** #665 notes that how management plane configuration should reach Azurite is still undecided. `--accountConfigFile` / `--accountConfig` is one answer; if the team prefers a different mechanism, the surface is contained to `src/common/account/`, the three environment classes and `BlobServerFactory`, and can be swapped without touching the versioning logic. Happy to change it.
- The account configuration store lives in `src/common/account/` so queue and table can share it, but the blob service owns its lifecycle for now as the only consumer. When another service reads it, ownership needs to move to the entry point and the instance be shared — two LokiJS instances autosaving one file would corrupt it. This is recorded on the field in `BlobServer` and in the design doc.
- Not implemented and out of scope, documented rather than partially emulated: blob version SAS (`sr=bv` and the `x` permission), soft-delete interactions, blob expiration, change feed, object replication, and `versionid` on Get Block List / Get Page Ranges (the swagger does not define it on either).


gavin-thompson-postman:
@microsoft-github-policy-service agree company="Postman"

gavin-thompson-postman:
Closing as hopefully https://github.com/Azure/Azurite/pull/2735 will be resolve this

PR #2749, opened 2026-08-25
Implement Put Blob From URL

### LINKED ISSUE #2353: Support Put Blob From URL
This is a feature request.

gaul:
Put Blob From URL answered 501, so a client copying a blob in one request had to fall back to the asynchronous Copy Blob and then wait out a copy it had asked to have done by the time it was answered.  Read the source the way Put Block From URL already does -- a loopback self-request pinned to the address and port this server is bound to, carrying the caller's path, query and source conditions -- and commit what comes back as a new block blob.  The two operations now share that fetch, so authentication, conditions, and the refusal to decompress a source that merely declares an encoding are answered the same way for both, in one place.

What the destination ends up with follows the operation's contract. The source's standard properties are copied unless x-ms-copy-source-blob-properties says otherwise, and a blob content header on the request sets that one property either way.  Metadata answers to its own rule: named on the request it replaces the source's, named nowhere it is copied.  Tags are the request's, or the source's under x-ms-copy-source-tag-option: COPY, which reads them over the same authorized path the content came over -- the extra Get Blob Tags call against the source that the operation is documented to make, so a URL allowed to read the source but not its tags is refused rather than obeyed.  Asking for both at once is refused as it already is for Copy Blob From URL.

The operation writes a blob, so it joins Put Blob in the shared access signature tables and in the rule that an existing destination demands write permission; an operation missing from those tables fails the request with an internal error rather than a permission one.

References #2681.


ChangeLog.md
```diff
@@ -61,6 +61,7 @@ Blob:
...
 - Added CRC-64/NVME transactional checksum support for `StageBlock`, `PutBlock`, `PutBlob`, `AppendBlock`, and `PutPage` (`x-ms-content-crc64`).
 - Harden transactional checksum validation for `PutBlob`, `StageBlock`, `AppendBlock`, and `PutPage`: unified MD5/CRC64 validation logic with accurate `InvalidMd5`/`InvalidHeaderValue` (malformed) and `Md5Mismatch`/`Crc64Mismatch` (mismatch) errors, matching real Azure semantics verified against live.
+- Implement `PutBlobFromUrl` (`Put Blob From URL`), which previously returned 501. The source is fetched over loopback, as `PutBlockFromURL` already does, so that SAS authentication and the `x-ms-source-if-*` conditions are enforced by the existing download path. Standard blob properties are copied from the source unless `x-ms-copy-source-blob-properties` is false, request blob content headers override them either way, request metadata replaces the source's rather than adding to it, and `x-ms-copy-source-tag-option: COPY` reads the source's tags over that same authorized path. As with `CopyBlobFromURL`, only sources on the same Azurite instance are supported.
```
jainakanksha-msft: @gaul , let's add this in upcoming release section of new release.

gaul: Done.

src/blob/handlers/BlockBlobHandler.ts
```diff
@@ -208,9 +207,222 @@ export default class BlockBlobHandler
...
+    let calculatedContentMD5: Uint8Array | undefined;
+    try {
+      ({ md5: calculatedContentMD5 } =
+        await computeAndValidateTransactionalChecksums(
+          stream,
+          {
+            md5:
+              options.sourceContentMD5 ??
+              (options.blobHTTPHeaders || {}).blobContentMD5
+          },
```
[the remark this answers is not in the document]
gaul: Done in 95e01316. Content-MD5 now joins the validation precedence (x-ms-source-content-md5, then x-ms-blob-content-md5, then Content-MD5, the order Put Blob uses), and a malformed value in any of the three is rejected with InvalidMd5 before the source is fetched. The current REST reference does not list Content-MD5 among this operation's request headers, although the swagger carries it and the page's response-header and CRC64 rows refer to it, so this follows Put Blob's semantics rather than a live check.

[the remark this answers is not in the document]
gaul: Done.

src/blob/authentication/BlobSASAuthenticator.ts
```diff
@@ -418,6 +418,7 @@ export default class BlobSASAuthenticator implements IAuthenticator {
     // If copy destination blob exists, then permission must be Write only
     if (
       operation === Operation.BlockBlob_Upload ||
+      operation === Operation.BlockBlob_PutBlobFromUrl ||
       operation === Operation.PageBlob_Create ||
       operation === Operation.AppendBlob_Create ||
       operation === Operation.Blob_StartCopyFromURL ||
```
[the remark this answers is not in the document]
gaul: The current behavior matches the documented permission split for this operation. The Put Blob From URL reference's SAS tables grant creating a new block blob to Create (c) or Write (w) and overwriting an existing one to Write (w) alone, the same as Put Blob. The permission table accepts either, and the exists-then-Write check refuses a c-only token that would overwrite. Requiring Create for a new blob would reject w-only tokens that Azure accepts. f0acc0ee adds SAS tests pinning all four cells of that table for account and blob SAS.

src/blob/authentication/OperationAccountSASPermission.ts
```diff
@@ -363,6 +363,16 @@ OPERATION_ACCOUNT_SAS_PERMISSIONS.set(
   )
 );
 
+OPERATION_ACCOUNT_SAS_PERMISSIONS.set(
+  Operation.BlockBlob_PutBlobFromUrl,
+  new OperationAccountSASPermission(
+    AccountSASService.Blob,
+    AccountSASResourceType.Object,
+    // Create permission is only available for nonexistent block blob. Handle this scenario separately
+    AccountSASPermission.Write + AccountSASPermission.Create
```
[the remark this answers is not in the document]
gaul: Reworded in 65dc5e9e. The entry now says Create or Write creates the blob, and overwriting an existing one takes Write alone, which the authenticator checks separately.

src/blob/authentication/OperationBlobSASPermission.ts
```diff
@@ -258,6 +258,13 @@ OPERATION_BLOB_SAS_BLOB_PERMISSIONS.set(
     BlobSASPermission.Write + BlobSASPermission.Create
   )
 );
+OPERATION_BLOB_SAS_BLOB_PERMISSIONS.set(
+  Operation.BlockBlob_PutBlobFromUrl,
+  // TODO: When destination blob doesn't exist, needs create permission
+  new OperationBlobSASPermission(
+    BlobSASPermission.Write + BlobSASPermission.Create
+  )
```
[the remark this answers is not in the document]
gaul: Reworded both entries in 65dc5e9e the same way. Create or Write creates the blob, and overwriting an existing one takes Write alone, which the authenticator checks separately.

jainakanksha-msft:
@gaul ,
x-ms-content-crc64 is neither accepted nor returned. The REST contract requires validation when supplied and says successful responses always include computed CRC64. The handler only selects MD5 inputs and computes MD5. The checked-in Swagger also omits CRC64 for this operation, so generated models/mappers/specification need updating too.
could you please add necessary handling for crc64 as well.

gaul:
@jainakanksha-msft Added in 6951d71e.

Put Blob From URL now accepts `x-ms-content-crc64` and checks it against the copied content: a mismatch is rejected as `Crc64Mismatch`, a value shorter than 8 bytes as `InvalidHeaderValue`, and a CRC64 sent alongside an MD5 header as `BothCrc64AndMd5HeaderPresent`, as Put Blob does. Every successful response carries the CRC64 computed over the copied content.

The swagger gains the `ContentCrc64` parameter and the `x-ms-content-crc64` 201 header for `BlockBlob_PutBlobFromUrl`, and the generated models, mapper and operation spec were updated to match. The response header mirrors the upstream client swagger; the request parameter is not in the client swagger, so it is recorded as change 18 in swagger/blob.md.

Tests cover the returned header, a matching value, a mismatch, a short value, and CRC64 combined with either `Content-MD5` or `x-ms-source-content-md5`.

tests/blob/sas.test.ts
```diff
@@ -1688,6 +1776,86 @@ describe("Shared Access Signature (SAS) authentication", () => {
     await blob2SAS.syncCopyFromURL(blob1.url);
   });
 
+  it("Put blob from URL with write permission in blob SAS should create and override a blob @loki @sql", async () => {
```
jainakanksha-msft: @gaul These tests are named "blob SAS," but they construct the credential with `containerClient.generateSasUrl()` and `ContainerSASPermissions`, so they exercise a container-scoped SAS (`sr=c`). Could we rename them accordingly and add true blob-scoped coverage using `blockBlobClient.generateSasUrl()` with `BlobSASPermissions`? Please also add a negative creation case for each independent authorization path: account SAS, container SAS, and blob SAS with neither `c` nor `w` (for example, `r` only) should fail with `403 AuthorizationPermissionMismatch`. The current tests cover `c`/`w` success and `c`-only overwrite denial, but do not pin rejection when neither required permission is present.

src/blob/handlers/BlockBlobHandler.ts
```diff
@@ -208,9 +208,248 @@ export default class BlockBlobHandler
...
+    // outranked MD5.
+    const blobHTTPHeaders = options.blobHTTPHeaders || {};
+    const requestMD5s = [
+      options.sourceContentMD5,
+      blobHTTPHeaders.blobContentMD5,
+      options.transactionalContentMD5
+    ];
+    const expectedContentMD5 = requestMD5s.find((md5) => md5 !== undefined);
+    const expectedContentCRC64 = options.transactionalContentCrc64;
+    if (
```
jainakanksha-msft: @gaul Could we move this pre-fetch checksum-header validation into the shared checksum utilities? `computeAndValidateTransactionalChecksums()` already owns the selected MD5/CRC64 mutual-exclusion and shape checks, while this block and `stageBlockFromURL()` maintain additional versions of the same rules. A synchronous helper could validate all supplied MD5 candidates plus CRC64 (accepting the caller-specific CRC64 header name), return the selected expected pair, and be reused by the compute helper and both URL handlers. That would preserve early rejection before fetching the source while keeping error semantics in one place. This is a maintainability suggestion rather than a functional blocker.

src/blob/handlers/BlockBlobHandler.ts
```diff
@@ -208,9 +208,248 @@ export default class BlockBlobHandler
...
+          { md5: true, crc64: true }
+        ));
+    } finally {
+      (stream as Readable).destroy?.();
+    }
+
+    // COPY reads the source's tags over the same authorized path the content
+    // came over, so a source that the caller may read but not tag refuses
+    // the copy rather than leaking them.
+    const blobTags = copySourceTags
```
jainakanksha-msft, line 329: @gaul Azure requires the caller to satisfy `Set Blob Tags` authorization when `x-ms-tags` sets destination tags, and `x-ms-copy-source-tag-option: COPY` likewise requires Set Blob Tags permission on the destination. This handler persists those tags, but the new account/blob/container SAS mappings only require `c` or `w`, so a destination SAS without `t` can currently write tags. Could we conditionally require `t` when this operation sets or copies destination tags, with account-, container-, and blob-scoped SAS tests? Requests that do not write tags should continue to require only the documented `c`/`w` permissions.

tests/blob/sas.test.ts
gaul: Done. The container-signed tests are now named for the container SAS, blob-scoped tests using `blockBlobClient.generateSasUrl` with `BlobSASPermissions` sit beside them, and each of account, container and blob SAS has a read-only negative asserting 403 AuthorizationPermissionMismatch.

src/blob/handlers/BlockBlobHandler.ts
gaul: Done. `validateTransactionalChecksumHeaders` in utils.ts now owns the mutual-exclusion and shape checks. It takes the MD5 candidates in precedence order and the CRC64 header name to report, and is called by `computeAndValidateTransactionalChecksums`, `putBlobFromUrl` and `stageBlockFromURL`, so early rejection stays and the rules live in one place.

src/blob/handlers/BlockBlobHandler.ts
gaul: Done. Both SAS authenticators now require Tag for this operation when `x-ms-tags` is present or `x-ms-copy-source-tag-option` is COPY, and leave requests that set no tags to Create or Write. Tests cover account, container and blob SAS. Note the same doc sentence applies to Put Blob and Copy Blob From URL, which Azurite does not enforce today; I left those as they were, since changing them is outside this PR.

